PR #35, opened 2017-10-28
use WDIOC_KEEPALIVE from golang.org/x/sys/unix

tklauser:
...now that it was added there.


stapelberg:
Thanks, you raced me to it :)

PR #36, opened 2018-03-26
dhcp: simplify Utsname conversion

tklauser:
Utsname from golang.org/x/sys/unix contains byte array members instead of
int8/uint8. This allows to simplify the NULL byte indexing in
addHostname.


stapelberg:
Thanks!

PR #42, opened 2018-10-05
reboot: use KexecFileLoad and KEXEC_* from golang.org/x/sys/unix

tklauser:
The KexecFileLoad syscall wrapper and the KEXEC_* constants were added
to golang.org/x/sys/unix. Use them instead of open-coding them.


stapelberg:
Neat, thanks! I’ll give this a shot later today.

stapelberg:
Seems to work fine, thanks for the cleanup!

PR #53, opened 2020-04-28
Added proof of concept TLS support (#7)

janisstreib:
If there is a certificate and key in `/etc/ssl/gokrazy-web.{pem,key.pem}`, the web
interface listens and redirects to https.

This patch comes along with other patches in the [internal](https://github.com/gokrazy/internal/pull/4) and [tools](https://github.com/gokrazy/tools/pull/14) repositories.

### DIFF
--- a/gokrazy.go
+++ b/gokrazy.go
@@ -20,9 +20,9 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/gokrazy/gokrazy/internal/iface"
-
 	"golang.org/x/sys/unix"
+
+	"github.com/gokrazy/gokrazy/internal/iface"
 )
 
 var (
@@ -68,24 +68,28 @@ func watchdog() {
 }
 
 func setupTLS() error {
-	if _, err := os.Stat("/etc/ssl/gokrazy-web.pem"); !os.IsNotExist(err) {
-		cert, err := tls.LoadX509KeyPair("/etc/ssl/gokrazy-web.pem", "/etc/ssl/gokrazy-web.key.pem")
-		if err != nil {
-			return fmt.Errorf("failed loading certificate: %v", err)
-		}
-		useTLS = true
-		// See https://cipherlist.eu/
-		tlsConfig = &tls.Config{
-			Certificates:             []tls.Certificate{cert},
-			MinVersion:               tls.VersionTLS12,
-			CurvePreferences:         []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
-			PreferServerCipherSuites: true,
-			CipherSuites: []uint16{
-				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-				tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
-			},
-		}
+	if _, err := os.Stat("/etc/ssl/gokrazy-web.pem"); os.IsNotExist(err) {
+		return nil // Nothing to set up
 	}
+	cert, err := tls.LoadX509KeyPair("/etc/ssl/gokrazy-web.pem", "/etc/ssl/gokrazy-web.key.pem")
+	if err != nil {
+		return fmt.Errorf("failed loading certificate: %v", err)
+	}
+	useTLS = true
+	tlsConfig = &tls.Config{
+		Certificates:             []tls.Certificate{cert},
+		MinVersion:               tls.VersionTLS12,
+		CurvePreferences:         []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256},
+		PreferServerCipherSuites: true,
+		CipherSuites: []uint16{
+			// required for http/2
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+			// See https://cipherlist.eu/
+			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
+		},
+	}
+	return nil
 }
 
 // Boot configures basic system settings. More specifically, it:
@@ -141,8 +145,8 @@ func Boot(userBuildTimestamp string) error {
 	}
 
 	initRemoteSyslog()
-    if err := setupTLS(); err != nil {
-    	return err
+	if err := setupTLS(); err != nil {
+		return err
 	}
 
 	return nil
--- a/gokrazy.go
+++ b/gokrazy.go
@@ -8,6 +8,7 @@
 package gokrazy
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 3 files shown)


gokrazy.go
```diff
@@ -118,6 +121,38 @@ func Boot(userBuildTimestamp string) error {
...
 
+	if _, err := os.Stat("/etc/ssl/web.pem"); !os.IsNotExist(err) {
+		cert, err := tls.LoadX509KeyPair("/etc/ssl/web.pem", "/etc/ssl/web_key.pem")
+		if err != nil {
+			return fmt.Errorf("failed loading certificate: %v", err)
+		} else {
+			useTLS = true
+			tlsConfig = &tls.Config{
+				Certificates: []tls.Certificate{cert},
+				// Let's be paranoid!
```
stapelberg, line 132: Should we follow the recommendations on https://cipherlist.eu/ for a Go server instead? (and reference that in the comment so that it can be updated)

gokrazy.go
```diff
@@ -118,6 +121,38 @@ func Boot(userBuildTimestamp string) error {
 
 	initRemoteSyslog()
 
+	if _, err := os.Stat("/etc/ssl/web.pem"); !os.IsNotExist(err) {
+		cert, err := tls.LoadX509KeyPair("/etc/ssl/web.pem", "/etc/ssl/web_key.pem")
+		if err != nil {
+			return fmt.Errorf("failed loading certificate: %v", err)
+		} else {
```
stapelberg, line 128: Kill this unnecessary else branch—you already return from the if branch :)

listeners.go
```diff
@@ -128,16 +129,27 @@ func PublicInterfaceAddrs() ([]string, error) {
...
+}
+
 var (
-	listeners   = make(map[string]*http.Server)
+	listeners   = make(map[string][]*krazyServer)
 	listenersMu sync.Mutex
 )
 
-func updateListeners(port string) error {
+// tlsEnbaled: Indicate, if public or private addresses should be used
```
stapelberg, line 143: typo: tlsEnabled

stapelberg:
Thanks for sending these PRs!

gokrazy.go
```diff
@@ -28,6 +29,8 @@ var (
 	buildTimestamp = "uninitialized"
 	httpPassword   string
 	hostname       string
+	tlsConfig      *tls.Config = nil
```
stapelberg, line 32: No need for the `= nil`, variables are initialized to their zero value by default.

gokrazy.go
```diff
@@ -118,6 +121,37 @@ func Boot(userBuildTimestamp string) error {
 
 	initRemoteSyslog()
 
+	if _, err := os.Stat("/etc/ssl/web.pem"); !os.IsNotExist(err) {
```
stapelberg, line 124: Can you factor this into a separate function please? Maybe `setupTLS()`

listeners.go
```diff
@@ -128,16 +129,27 @@ func PublicInterfaceAddrs() ([]string, error) {
...
+}
+
 var (
-	listeners   = make(map[string]*http.Server)
+	listeners   = make(map[string][]*krazyServer)
 	listenersMu sync.Mutex
 )
 
-func updateListeners(port string) error {
+// tlsEnabled: Indicate, if public or private addresses should be used
```
stapelberg, line 143: I wonder whether private/public address selection should really be tied to TLS being enabled.

Of course, when TLS is not enabled, HTTP should not be available on public addresses.
But, just because TLS is enabled, doesn’t mean I necessarily want to serve HTTP on public addresses.
Think of the unencrypted WiFi use-case: updating should still not necessarily be possible via the WAN.

Maybe we need another knob? It might be time to introduce a JSON config file with parameters for the gokrazy web interface. To cover the port-related TODO above, maybe we allow specifying the listen address.

janisstreib: I'd prefer the latter from a developer's perspective as well as from user's/administrator's perspective. For example: I live in a dormitory with a big routed network with private legacy IPs. Just allowing "private" addresses doesn't make a difference for me, as the service would be still reachable for about 40k people (including the university).

gokrazy.go
```diff
@@ -64,6 +67,27 @@ func watchdog() {
 	}
 }
 
+func setupTLS() error {
+	if _, err := os.Stat("/etc/ssl/gokrazy-web.pem"); !os.IsNotExist(err) {
```
stapelberg, line 71: Invert this conditional and use early return:

```
if _, err := os.Stat("/etc/ssl/gokrazy-web.pem"); os.IsNotExist(err) {
  return // nothing to set up
}
```

gokrazy.go
```diff
@@ -117,10 +141,25 @@ func Boot(userBuildTimestamp string) error {
 	}
 
 	initRemoteSyslog()
+    if err := setupTLS(); err != nil {
```
stapelberg, line 144: Looks like you didn’t run gofmt on your code? In general, I’d recommend to set up your editor to automatically run `goimports` on save: https://godoc.org/golang.org/x/tools/cmd/goimports

httpsredirect.go
```diff
@@ -0,0 +1,9 @@
+package gokrazy
+
+import "net/http"
+
+// TODO: Configurable https-port
+func httpsRedirect(w http.ResponseWriter, r *http.Request) {
+	newURI := "https://" + r.Host + r.URL.String()
```
stapelberg, line 7: I think it’d be slightly cleaner/more robust to modify the r.URL fields like so:
```
r.URL.Host = r.Host
r.URL.Scheme = "https"
newURI := r.URL.String()
```

listeners.go
```diff
@@ -146,25 +158,55 @@ func updateListeners(port string) error {
...
+			}
+			if cont {
+				continue
+			}
 		}
 		addr := net.JoinHostPort(host, port)
 		ln, err := net.Listen("tcp", addr)
 		if err != nil {
 			log.Println(err) // err includes enough context
+			log.Println("(error on listen)")
```
stapelberg, line 179: So apparently err didn’t include enough context after all :)

Let’s make this `log.Printf("listen: %v", err)`, then?

listeners.go
stapelberg: > I'd prefer the latter

Sorry, when you say “the latter”, what do you mean?



> I live in a dormitory with a big routed network with private legacy IPs. Just allowing "private" addresses doesn't make a difference for me, as the service would be still reachable for about 40k people (including the university).

Right, but it also doesn’t hurt, right? I.e., you don’t necessarily need to use a public address, you can just keep using a private address (even though in your particular network private isn’t as private as elsewhere)?

janisstreib: > > I'd prefer the latter
> 
> Sorry, when you say “the latter”, what do you mean?

Allowing the user to specify the listen address(es). In my opinion, this would make the behavior of the server more "defined".
Should "listen private only" still be the fallback, if not defined otherwise or if the address is not configured on any interface?

gokrazy.go
janisstreib: Hm, I actually do this as pre-commit hook. I don't know what happened here.

PR #64, opened 2020-06-29
website: add consrv to showcase

mdlayher:
/cc @stapelberg


website/content/showcase.html
```diff
@@ -49,6 +49,11 @@ <h3>Written for gokrazy</h3>
 figurines for playing music stored on the device or directly from Spotify.
 </li>
 </ul>
```
stapelberg, line 51: Your entry should go before this `</ul>`

mdlayher: Ha, been a while since I've done HTML. Done.

mdlayher:
Done and done!

Issue #68, opened 2020-09-03
Firmware/BIOS information


CodeZombieCH:
I'm working on a PR to show firmware/BIOS information (in particular the version number) of my PC Engines router7 in the header of the web UI. With the recent addition of kernel information, I feel like adding an addition row with firmware information grows the header beyond a reasonable height. I came up with three options to display the firmware information in the header. Please let me know which one you would accept as a PR :-) 

## Option 1:
![compact](https://user-images.githubusercontent.com/729017/92144747-71418800-ee17-11ea-8570-438d1c18b4d1.png)
Compact version, where firmware information is added to the model information

## Option 2:
![single-column](https://user-images.githubusercontent.com/729017/92144756-74d50f00-ee17-11ea-80c8-8106ab88d9c2.png)
Current single column approach with detailed firmware

## Option 3:
![multiple-columns](https://user-images.githubusercontent.com/729017/92144750-730b4b80-ee17-11ea-8ef6-fa092517752f.png)
Multiple columns with detailed firmware information with very small height

stapelberg:
Thanks for the detailed suggestion!

I prefer option 3: it’s the most compact, and given the gokrazy web interface’s target audience of developers, I think most viewers will be on a screen that’s wider than high.

CodeZombieCH:
Thanks for the feedback. I forgot to mention that option 3 is my preferred option as well. I will update my bootstrap skills and create a PR within the next days.

PR #69, opened 2020-09-16
Added firmware information to web interface header

CodeZombieCH:
This PR implements #68 (Option 3). It also includes some minor styling improvements of the web interface header.

Unfortunately I was not able to find out how to retrieve the firmware version on the raspi3b. Any support on this is more than welcome.

## Testing

This PR has been tested on the following devices:

platform | expected result | test result
-- | -- | --
apu2c4 | shows correct firmware version  | successfully tested
raspi3b | no firmware information | successfully tested
raspi3b+ | no firmware information | not available due to lack of hardware
raspi4b | no firmware information | not available due to lack of hardware

## Follow up

While working on this feature I noticed that the apu2c4|raspi3b(+) firmware version and the raspi4b EEPROM version describe the same type of information (at least to my understanding of what a firmware is). As such, we could think about merging the two firmware information retrieval strategies into a single one with conditionals. As of now, on the raspi4b the firmware version is expected to be empty and the EEPROM version to be populated, so the header looks already pretty much the same except for the different wording.


status.go
```diff
@@ -79,6 +79,18 @@ func Model() string {
 	return strings.TrimSpace(model)
 }
 
+// Firmware returns a firmware description for the PC Engines apu2 (and other PCs), e.g. “coreboot v4.11.0.6 (04/26/2020)”.
+func Firmware() string {
+	vendor := mustReadFile0("/sys/class/dmi/id/bios_vendor")
+	version := mustReadFile0("/sys/class/dmi/id/bios_version")
+	date := mustReadFile0("/sys/class/dmi/id/bios_date")
+	if vendor == "" || version == "" || date == "" {
```
stapelberg, line 87: Instead of not displaying anything when any part is missing, I think it might be nice to display as much as we can. That way, the code works better on platforms we can’t test yet.

stapelberg:
> Unfortunately I was not able to find out how to retrieve the firmware version on the raspi3b. Any support on this is more than welcome.

If it is available, it’d need to be in the device tree.

If it’s not in there, maybe the firmware doesn’t actually expose a version.

status.go
CodeZombieCH: Fair point, fully agreed. I will change this accordingly

CodeZombieCH:
> > Unfortunately I was not able to find out how to retrieve the firmware version on the raspi3b. Any support on this is more than welcome.
> 
> If it is available, it’d need to be in the device tree.
> 
> If it’s not in there, maybe the firmware doesn’t actually expose a version.

If I remember correctly I already searched the device tree for firmware information. I will have a second look...

CodeZombieCH:
> > Unfortunately I was not able to find out how to retrieve the firmware version on the raspi3b. Any support on this is more than welcome.
> 
> If it is available, it’d need to be in the device tree.
> 
> If it’s not in there, maybe the firmware doesn’t actually expose a version.

I now had the time to investigate more about the Raspberry Pi firmware version:

Coming from the router7 setup, what I intended to describe by the "firmware" was the bootloader and it's version. In case of my router7 running on the apu2e4, this is "coreboot". On the Raspberry Pi, the bootloader seems to be built into the GPU and is non-updateable (see https://wiki.ubuntu.com/ARM/RaspberryPi#Create_a_partition_with_pi_firmware.2Fbootloader_files). If I understand correctly, both bootloader then start loading a boot payload. In case of the apu2e4, the payload is a version of SeaBIOS stored in an on-board flash chip. In case of the Raspberry Pi below Model 4, the payload is stored in the `bootcode.bin` file on the boot partition. Starting with Model 4, the payload is stored on the EEPROM.

As a consequence, I think my choice of the word "firmware" might have been a bad decision. We might be better of using the word "bootloader" and show this information only on the apu2e4, as the bootloader on the Raspberry Pi will not change. What changes thought is the boot payload, which has already been taken care of with the "EEPROM" information.

Please let me know about your thoughts on this.

status.go
CodeZombieCH: See https://github.com/gokrazy/gokrazy/pull/69/files#diff-9ebb930cab0da6fd9029098cf29ed2d6R82-R99

assets/header.tmpl
```diff
@@ -39,38 +39,45 @@ table {
...
+          {{ if (ne .Firmware "") }}
+          <tr>
+            <th>firmware</th>
+            <td>{{ .Firmware }}</td>
+          </tr>
+          {{ end }}
+          {{ if .EEPROM }}
+          <tr>
+            <th>EEPROM<br>(SHA256)</th>
+            <td>{{ shortenSHA256 .EEPROM.PieepromSHA256 }}<br>{{ shortenSHA256 .EEPROM.VL805SHA256 }}</td>
```
stapelberg, line 75: Can you add “(EEPROM)” and “(VL805)” respectively, so that it’s more clear what these digests are referring to please?

stapelberg (commented):
Sorry for the late reply.

Could you rebase your PR onto the current `master` branch and squash commits into one logical commit for a clean git history please?

stapelberg:
> As a consequence, I think my choice of the word "firmware" might have been a bad decision. We might be better of using the word "bootloader" and show this information only on the apu2e4, as the bootloader on the Raspberry Pi will not change. What changes thought is the boot payload, which has already been taken care of with the "EEPROM" information.
> 
> Please let me know about your thoughts on this.

Sounds good to me!

damdo:
I think this would be handy, glancing at the UI and finding out the firmware version seems useful and would have helped me also recently figuring out a certain firmware bump was breaking my build.
Are you still interested in doing so @CodeZombieCH ?
Thanks

slashformotion:
Is anyone interested in rebasing this ?

stapelberg:
I submitted commit c243e28f616da8d9bc41c54fd34fc31f45e24d97 to get this PR off our plate. Here’s how it looks like on a PC Engines APU2:

2026-09-05-gokrazy-firmware

Issue #74, opened 2021-01-03
Allow custom ports for gokrazy ui/api


andig:
## Platform

n/a

## Observed behavior

gokrazy uses http:80 for it's own purposes. I'd like to use that for the bundled application.

## Expected behavior

Specify custom port for created package, to be reused by updater.

Would be happy to work on a PR.

stapelberg:
I think this is a reasonable feature request, just not entirely sure yet about the configuration surface.

Maybe another file (`port`) in `~/.config/gokrazy`?

andig:
I've noticed `~/.config` in the quickstart but at that time wasn't sure if that's on the host (it is). I don't have `~/.config` on my mac after building an image- when is it generated for the default password?

For the port I'd say that it is image-specific if it needs be moved or not. Would it hurt to allow a `gokrazy` folder in side cwd that- if exists- takes precedence over `~/.config/gokrazy`?

Simple file with just `80,443` as content? Injected into build using build vars?

Merovius:
I think you really want [os.UserConfigDir](https://golang.org/pkg/os/#UserConfigDir), whatever that may end up being.

stapelberg:
> For the port I'd say that it is image-specific if it needs be moved or not

gokrazy’s configuration is centered around the concept of a hostname.

> I think you really want [os.UserConfigDir](https://golang.org/pkg/os/#UserConfigDir), whatever that may end up being.

We do use os.UserConfigDir, see https://github.com/gokrazy/internal/blob/master/config/config.go

> Simple file with just `80,443` as content?

I would suggest one port number per file. Perhaps we could name them `port-http` and `port-https`.

> Injected into build using build vars?

I don’t think any build vars are necessary? What did you have in mind?

fawick:
>  I don’t think any build vars are necessary? What did you have in mind?

The possibility to state additional Go build tags and or linker flags for setting symbol values (version info) would be nice.

andig:
> gokrazy’s configuration is centered around the concept of a hostname.

Got it.

> The possibility to state additional Go build tags and or linker flags for setting symbol values (version info) would be nice.

Instead of modifying the host program it would be nice to use the right build tags during packaging (`build-tags` file?).

andig:
There are two todos here:

- [ ] implement runtime: https://github.com/gokrazy/gokrazy/pull/77
- [ ] implement host tooling: help wanted

[...]

PR #77, opened 2021-01-06
Load runtime port configuration from http(s)-port file

andig:
This PR provides the runtime configuration for https://github.com/gokrazy/gokrazy/issues/74. 
The host part for creating the files from `gokr-packager` is still missing.

### DIFF
--- a/gokrazy.go
+++ b/gokrazy.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


gokrazy.go
```diff
@@ -104,6 +105,19 @@ func setupTLS() error {
 	return nil
 }
 
+// readFile reads configuration files from /perm /etc or / and returns trimmed content as string
+func readFile(fileName string) (string, error) {
```
stapelberg, line 109: Can you name this readConfigFile to be a little more descriptive please?

gokrazy.go
```diff
@@ -223,7 +230,17 @@ func Supervise(commands []*exec.Cmd) error {
 		return err
 	}
 
-	if err := updateListenerPairs("80", "443", useTLS, tlsConfig); err != nil {
+	httpPort, err := readFile("http-port.txt")
+	if _, numeric := strconv.Atoi(httpPort); err != nil || numeric != nil {
+		httpPort = "80"
```
stapelberg, line 235: Here and below: please log an error message when the file is present but malformed, so that users know what the problem is.

andig: Split into function as the validation with logging is more than 2 lines

andig:
I can confirm that with https://github.com/gokrazy/tools/commit/90d0f0f832e71126a70529ec51bd5dd6ef663e18 applied I can curl into the port specified with `http_port`.

If you want to merge and upgrade `gokrazy/tools` to use it, please upgrade `x/sys` there to latest as well or it won't build on OSX.

stapelberg (approved):

Issue #25, opened 2017-04-19
Auto-updating from gokrazy


Merovius:
As discussed offline, it would be great, if gokrazy would support full auto-updates (driven by gokrazy itself).
It would probably require shipping a copy of the go tool and the packer, regularly checking for updates via the github API, fetching a tarball of the content (as we won't have git/mercurial/svn on the rpi) if there is one, rebuilding the image and rebooting. Most of this can probably be built as a service, though some changes to the packer are required: a) The packer needs to add a released binary of the go tool (we probably don't want to do a full recompilation?) and b) it needs to add meta-information about the build, so that the packer running on gokrazy can rebuild it.

It's probably also worth it to consider user-packages for auto-updates, though the process is probably going to be different. Most third-party things probably shouldn't be built from HEAD, so either we need to rely on some kind of versioning scheme to decide what to update to, or rely on github releases and download binaries.
Updated versions should first be written to and started from /tmp, to make sure that they start correctly. Afterwards, the packer needs to build a new image containing the updated binary.

I think in the end we'd want auto-updates both for gokrazy itself and the user-installed services, to have fully self-updating appliances.

stapelberg:
> The packer needs to add a released binary of the go tool (we probably don't want to do a full recompilation?)

I think it’d be nice if the auto-update mechanism lived in a separate package, e.g. github.com/gokrazy/selfdestruct. We’d need to find a way to package the go tool (and gokr-packer) so that `go install` installs their binaries, then gokr-packer would just pick it up like any other package. Merely copying the host’s go tool and gokr-packer would not work in many cases: the host binaries are not compiled with `CGO_ENABLED=0`. Thoughts?

> Most third-party things probably shouldn't be built from HEAD, so either we need to rely on some kind of versioning scheme to decide what to update to, or rely on github releases and download binaries.

Playing the version game is something we can’t win right now, I think. gokrazy intentionally uses the go tool for fetching and building Go packages, and I don’t feel compelled to change that currently (this may change when/if the dep tool becomes part of the toolchain). If people have more complicated needs for determining the version, they need to either:

* put that logic into some custom infrastructure and produce go gettable packages
* do “external” (as opposed to on-device) updates

FWIW, here’s the script I run via cron to automatically update my installation:

```sh
#!/bin/sh

export GOPATH=/home/michael/go
export PATH=/home/michael/go/bin:/home/michael/go-latest/bin:$PATH

go get -u github.com/gokrazy/kernel
[...]

PR #78, opened 2021-01-07
Don't redirect when localhost

andig:
Any requests with localhost as target (though not an IP address) will not be redirected. Helps with https://github.com/gokrazy/gokrazy/issues/25.

I'll send a separate PR for redirecting to the right http port (#79).

### DIFF
--- a/httpsredirect.go
+++ b/httpsredirect.go
@@ -8,7 +8,7 @@ import (
 func httpsRedirect(w http.ResponseWriter, r *http.Request) {
 	host, _, _ := net.SplitHostPort(r.RemoteAddr)
 	ip := net.ParseIP(host)
-	if IsInPrivateNet(ip) {
+	if ip.IsLoopback() {
 		http.DefaultServeMux.ServeHTTP(w, r)
 		return
 	}
--- a/httpsredirect.go
+++ b/httpsredirect.go
@@ -7,7 +7,12 @@ import (
 
 func httpsRedirect(w http.ResponseWriter, r *http.Request) {
 	// get hostname without port
-	ip := net.ParseIP(r.RemoteAddr)
+	host, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		host = r.RemoteAddr
+	}
+
+	ip := net.ParseIP(host)
 	if IsInPrivateNet(ip) {
 		http.DefaultServeMux.ServeHTTP(w, r)
 		return
--- a/httpsredirect.go
+++ b/httpsredirect.go
@@ -6,12 +6,7 @@ import (
 )
 
 func httpsRedirect(w http.ResponseWriter, r *http.Request) {
-	// get hostname without port
-	host, _, err := net.SplitHostPort(r.RemoteAddr)
-	if err != nil {
-		host = r.RemoteAddr
-	}
-
+	host, _, _ := net.SplitHostPort(r.RemoteAddr)
 	ip := net.ParseIP(host)
 	if IsInPrivateNet(ip) {
 		http.DefaultServeMux.ServeHTTP(w, r)
--- a/httpsredirect.go
+++ b/httpsredirect.go
@@ -1,10 +1,27 @@
 package gokrazy
 
-import "net/http"
+import (
+	"net/http"
+	"strings"
+)
 
 // TODO: Configurable https-port
-func httpsRedirect(w http.ResponseWriter, r *http.Request) {
-	r.URL.Host = r.Host
-	r.URL.Scheme = "https"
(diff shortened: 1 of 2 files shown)


httpsredirect.go
```diff
@@ -1,10 +1,27 @@
...
 
 // TODO: Configurable https-port
-func httpsRedirect(w http.ResponseWriter, r *http.Request) {
-	r.URL.Host = r.Host
-	r.URL.Scheme = "https"
-	http.Redirect(w, r, r.URL.String(), http.StatusFound) // Redirect to https
+func httpsRedirect(handler http.Handler) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// get hostname without port
+		target := r.URL.Host
```
stapelberg, line 12: Unfortunately, this approach isn’t safe: the “Host” HTTP header is under control by the attacker, not the server.

We need to look at https://golang.org/pkg/net/http/#Request RemoteAddr field. See the existing `IsInPrivateNet` function for how to compare IP addresses with networks, and then compare with just 127.0.0.0/8 and ::1/128.

listeners.go
```diff
@@ -180,9 +180,9 @@ func updateListeners(port string, tlsEnabled bool, tlsConfig *tls.Config) error
 		if tlsEnabled && tlsConfig == nil {
 			// "Redirect" server
 			srv = &http.Server{
-				Handler:   http.HandlerFunc(httpsRedirect),
 				TLSConfig: tlsConfig,
 			}
+			srv.Handler = http.HandlerFunc(httpsRedirect(srv.Handler))
```
stapelberg, line 185: The srv.Handler parameter will just be nil, no? Why not keep the line as it was and specify nil explicitly?

httpsredirect.go
```diff
@@ -1,10 +1,27 @@
 package gokrazy
 
-import "net/http"
+import (
+	"net/http"
+	"strings"
+)
 
 // TODO: Configurable https-port
```
stapelberg, line 8: Please remove this TODO while here :)

listeners.go
andig: Argh, I had expected that to be `http.DefaultHandlerMux`, you're right.

httpsredirect.go
```diff
@@ -1,27 +1,19 @@
...
-			handler.ServeHTTP(w, r)
-			return
-		}
-
-		r.URL.Host = r.Host
-		r.URL.Scheme = "https"
-		http.Redirect(w, r, r.URL.String(), http.StatusFound) // Redirect to https
+func httpsRedirect(w http.ResponseWriter, r *http.Request) {
+	// get hostname without port
+	ip := net.ParseIP(r.RemoteAddr)
```
stapelberg, line 10: r.RemoteAddr contains [ip]:port, AFAIR, so you’ll need to call https://golang.org/pkg/net/#SplitHostPort

httpsredirect.go
```diff
@@ -1,27 +1,19 @@
...
-		if target == "localhost" {
-			handler.ServeHTTP(w, r)
-			return
-		}
-
-		r.URL.Host = r.Host
-		r.URL.Scheme = "https"
-		http.Redirect(w, r, r.URL.String(), http.StatusFound) // Redirect to https
+func httpsRedirect(w http.ResponseWriter, r *http.Request) {
+	// get hostname without port
```
stapelberg, line 9: This comment is now obsolete (no hostnames involved)

httpsredirect.go
```diff
@@ -1,27 +1,19 @@
...
+	ip := net.ParseIP(r.RemoteAddr)
+	if IsInPrivateNet(ip) {
```
stapelberg, line 11: This is a bit too loose: we need a variant that not only checks the IP is private, but specifically that it’s IPv4/IPv6 localhost.

httpsredirect.go
andig: Removed the err check as well since:

> The HTTP server in this package sets RemoteAddr to an "IP:port" address before invoking a handler. This field is ignored by the HTTP client.

httpsredirect.go
andig: Could you grab this? Not sure what to do, sorry.

chlunde: What IP do you connect to?
If you connect to localhost I think this will work: https://golang.org/src/net/ip.go?s=3355:3385#L111

But if you connect to "the IP of eth0", like 192.168.10.10, then it's more difficult.

andig: Lets support this only for loopback, localhost is fine for the purpose.

stapelberg (approved):

PR #45, opened 2018-12-15
status API

Xe:
Creates two API calls to better integrate with my network and monitoring tools:

- /api/status: returns status information for the machine as a whole
- /api/service?path=/{gokrazy|user}/whatever: status for an individual service

These routes are protected behind HTTP basic authentication.

PR #82, opened 2021-01-14
Provide json status api

andig:
Replaces https://github.com/gokrazy/gokrazy/pull/45

### DIFF
--- a/status.go
+++ b/status.go
@@ -280,7 +280,7 @@ func initStatus(services []*service) {
 			}
 
 			w.Header().Set("Content-Type", "application/json")
-			_, _ = fmt.Fprint(w, string(b))
+			_, _ = w.Write(b)
 			return
 		}
 
@@ -370,7 +370,7 @@ func initStatus(services []*service) {
 			}
 
 			w.Header().Set("Content-Type", "application/json")
-			_, _ = fmt.Fprint(w, string(b))
+			_, _ = w.Write(b)
 			return
 		}
 
--- a/status.go
+++ b/status.go
@@ -171,10 +171,6 @@ func jsonRequested(r *http.Request) bool {
 	return strings.Contains(strings.ToLower(r.Header.Get("Content-type")), "application/json")
 }
 
-func jsonRequested(r *http.Request) bool {
-	return strings.Contains(strings.ToLower(r.Header.Get("Content-type")), "application/json")
-}
-
 func initStatus(services []*service) {
 	model := Model()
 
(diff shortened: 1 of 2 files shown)


supervise.go
```diff
@@ -197,6 +198,24 @@ func (s *service) setProcess(p *os.Process) {
...
+func (s *service) MarshalJSON() ([]byte, error) {
+	return json.Marshal(&struct {
+		Stopped   bool
+		StartTime time.Time
+		Attempt   uint64
+		Pid       int
+		Path      string
+		Args      []string
+	}{
+		Stopped:   s.stopped,
```
stapelberg, line 210: You’ll need to acquire stoppedMu, startedMu and processMu to safely access these fields.

Also, I think the attempt field is not currently used actually — I can only see usage of a local `attempt` variable.

stapelberg (changes requested):

status.go
```diff
@@ -264,6 +271,19 @@ func initStatus(services []*service) {
...
+
+		if jsonRequested(r) {
+			b, err := json.Marshal(svc)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+
+			w.Header().Set("Content-Type", "application/json")
+			_, _ = fmt.Fprint(w, string(b))
```
stapelberg, line 283: Here and below: don’t use `Fprint` without specifying a format. Instead, just use `w.Write(b)` for writing out JSON.

stapelberg (changes requested):

stapelberg (approved):

andig:
For sake of reference, here's the actual output:

    ❯ curl -v -H "content-type: application/json" http://gokrazy:xx@evcc:8080/status\?path\=/user/brokenglass
    * Server auth using Basic with user 'gokrazy'
    > GET /status?path=/user/brokenglass HTTP/1.1
    > Host: evcc:8080
    > User-Agent: curl/7.64.1
    > Accept: */*
    > content-type: application/json
    > 
    < HTTP/1.1 200 OK
    < Access-Control-Allow-Origin: *
    < Content-Type: application/json
    < Set-Cookie: gokrazy_xsrf=615554735; Expires=Mon, 18 Jan 2021 19:02:33 GMT; HttpOnly
    < Date: Sun, 17 Jan 2021 19:02:33 GMT
    < Content-Length: 131
    < 
    {"Stopped":true,"StartTime":"1970-01-01T01:00:21.98437546+01:00","Pid":175,"Path":"/user/brokenglass","Args":["/user/brokenglass"]}

PR #90, opened 2021-08-17
internal/iface: use unix.Ifreq API for interface flags ioctl

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

I'm working on a safer API for ifreqs and figured I might find some use for it in gokrazy! Arguably these calls could also be replaced with rtnetlink, but this helps for now.

I want to incorporate a sockaddr version too, but I'm still working on the API for that one. Will CC you on the CL.


stapelberg (approved):

stapelberg:
Thanks!

PR #93, opened 2021-09-15
Quote and separate the execeuted command in log

christian-heusel:
I had trouble understanding why the command (with args) I wanted to execute on gokrazy failed, because i could not see the diffrence between `['/user/echo', 'test 123']` and `['/user/echo', 'test', '123']`

Output (for both examples) before:
```
  2021/09/15 13:34:00 gokrazy: attempt 1, starting [/user/echo test 123]
```
Output now:
```
  2021/09/15 13:34:00 gokrazy: attempt 1, starting ['/user/echo', 'test', '123']
```

### DIFF
--- a/supervise.go
+++ b/supervise.go
@@ -296,8 +296,9 @@ func supervise(s *service) {
 			time.Sleep(1 * time.Second)
 			continue
 		}
-
-		l.Printf("gokrazy: attempt %d, starting %+v", attempt, s.cmd.Args)
+		// quote and separate the args so stderr shows what cmd is actually executed
+		argstr := "['" + strings.Join(s.cmd.Args, `', '`) + `']`
+		l.Printf("gokrazy: attempt %d, starting %s", attempt, argstr)
 		s.setStarted(time.Now())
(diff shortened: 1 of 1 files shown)


supervise.go
```diff
@@ -296,8 +296,9 @@ func supervise(s *service) {
 			time.Sleep(1 * time.Second)
 			continue
 		}
-
-		l.Printf("gokrazy: attempt %d, starting %+v", attempt, s.cmd.Args)
+		// quote and separate the args so stderr shows what cmd is actually executed
+		argstr := "['" + strings.Join(s.cmd.Args, `', '`) + `']`
+		l.Printf("gokrazy: attempt %d, starting %s", attempt, argstr)
```
stapelberg, line 301: Why not just change %+v to %q?

stapelberg (commented):
Thanks for your PR!

supervise.go
christian-heusel: Because I didnt know better :D See #95

PR #99, opened 2021-11-28
dhcp: allow builds on non-64-bit architectures

anupcshan:
Move build constraint to internal/iface which clearly shows the Linux dependency


stapelberg (approved):

stapelberg:
Thanks!

PR #100, opened 2021-12-12
status: live tail logs

anupcshan:
Create a `/log` endpoint for streaming stdout/stderr logs for processes.
Replace status page stdout/stderr to call into /log to stream logs and
live update page.

### DIFF
--- a/status.go
+++ b/status.go
@@ -171,6 +171,10 @@ func jsonRequested(r *http.Request) bool {
 	return strings.Contains(strings.ToLower(r.Header.Get("Content-type")), "application/json")
 }
 
+func eventStreamRequested(r *http.Request) bool {
+	return strings.Contains(strings.ToLower(r.Header.Get("Accept")), "text/event-stream")
+}
+
 var templates = template.Must(template.New("root").
 	Funcs(map[string]interface{}{
 		"shortenSHA256": func(hash string) string {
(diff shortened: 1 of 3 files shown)


anupcshan:
This is not an ideal way to implement live tailing (there are other options utilizing websockets), but it is quite simple and it does work. Happy to discuss other options.

supervise.go
```diff
@@ -107,7 +113,16 @@ func (lrb *lineRingBuffer) Write(b []byte) (int, error) {
 			break
 		}
 
-		lrb.r.Value = text[:idx]
+		line := text[:idx]
+		lrb.r.Value = line
+		for stream := range lrb.streams {
```
stapelberg, line 118: You need to hold the mutex when accessing `lrb.streams`, no?

When adding and deleting, you’re using the mutex correctly.

supervise.go
```diff
@@ -127,9 +142,35 @@ func (lrb *lineRingBuffer) Lines() []string {
 	return lines
 }
 
+// Stream generates a new channel which will stream any logged lines, including everything currently
+// in the ring buffer. Deregister the stream by calling the close function.
+func (lrb *lineRingBuffer) Stream() (<-chan string, func()) {
+	lrb.Lock()
+	defer lrb.Unlock()
+
+	// Need a chan that has at least len(ring) entries in it, otherwise populating it with existing
```
stapelberg, line 151: An alternative to sizing the stream channel is to make a copy of the lines and populate the existing contents from a separate goroutine. Then we don’t care how long it takes.

status.go
```diff
@@ -298,6 +302,52 @@ func initStatus(services []*service) {
...
+			stream, closeFunc = svc.Stdout.Stream()
+		case "stderr":
+			stream, closeFunc = svc.Stderr.Stream()
+		default:
+			http.Error(w, "stream not found", http.StatusNotFound)
+			return
+		}
+		defer closeFunc()
+
+		for line := range stream {
```
stapelberg, line 336: Can you tie this loop to the HTTP request context please? That way, we avoid goroutines piling up when the client has closed its connection but nothing is happening on the `stream` channel.

```go
ctx := r.Context()
for {
select {
case <-ctx.Done():
  return
case line := <-stream:
  // …
}
}
```

stapelberg (changes requested):
Very nice, thank you!

supervise.go
anupcshan: The lock is already taken in line 107 (at the beginning of `Write`)

supervise.go
anupcshan: Yeah, if we take a copy of lines and then async push them into the channel, its possible other lines can get pushed from `Write` while this is going on (and cause lines to be streamed out of order). Its super unlikely to happen, but something to keep in mind.

I'm skipping this alternative for now, but can go ahead with it if you feel this is a reasonable tradeoff.

status.go
anupcshan: Good call - done

anupcshan:
There is a slight difference in the log output from before - the current static log output adds one space to the beginning of each line. Is this something that needs to be added to the streaming one too?

stapelberg:
> There is a slight difference in the log output from before - the current static log output adds one space to the beginning of each line. Is this something that needs to be added to the streaming one too?

Sounds like an oversight to me. I don’t think we need to worry about this.

Issue #101, opened 2021-12-25
Support for Odroid HC2


anupcshan:
I have Gokrazy operating on an [Odroid HC2](https://www.hardkernel.com/shop/odroid-hc2-home-cloud-two/) and have been using it for 2-3 months now. I have a partially functional bootery setup going to continuously validate that the hardware works as expected and verify kernel upgrades (needs some change to how cmdline.txt gets updated for testboot - will bring that up once this issue is resolved).

There are a couple of issues that needed to be solved:
1. Support MBR-only partition table. This hardware has a weird booting scheme, loading bootloader blobs from [fixed locations on disk (sectors 1-2048)](https://wiki.odroid.com/odroid-xu4/software/partition_table#tab__odroid-xu341). That clobbers most of the GPT (everything after the primary header is overwritten) and I'm forced to switch back to MBR.
2. Support for writing/updating bootloader blobs at arbitrary positions on the bootable block device.
3. Add more entries to [kernelGlobs list](https://github.com/gokrazy/tools/blob/3bb6d5620329fbc2551333b93d2a9f0a3b3be648/cmd/gokr-packer/write.go#L175-L178).

The solutions I've come up with so far is:
1. Allow kernel package to optionally emit a YAML manifest that describes any requirements for this kernel/device (like MBR-only partition table, custom kernel globs and listing extra blobs). See [example](https://github.com/anupcshan/gokrazy-odroidxu4-kernel/blob/master/gokrazy-kernel-manifest.yaml).
2. [Support MBR-only partition table and parse that information from kernel manifest](https://github.com/anupcshan/tools/compare/3bb6d5620329fbc2551333b93d2a9f0a3b3be648...a95a1359834a4b14a46fc7989e3f519f308a92b2).
[...]

PR #102, opened 2021-12-29
update: support registering device-specific update handlers

anupcshan:
Useful to update devices which expect blobs at fixed locations

Related to #101


stapelberg (changes requested):
Thanks for your PR!

I think it’d be better to only offer a function that sets up the extra update handlers based on the device type, instead of offering a function that allows to set arbitrary update handlers.

E.g., how about `gokrazy.SetupDeviceSpecifics()`, taking a constant identifying the system (e.g. odroid)?

anupcshan:
> Thanks for your PR!
> 
> I think it’d be better to only offer a function that sets up the extra update handlers based on the device type, instead of offering a function that allows to set arbitrary update handlers.
> 
> E.g., how about `gokrazy.SetupDeviceSpecifics()`, taking a constant identifying the system (e.g. odroid)?

Hmmm, I'm trying to figure out where gokrazy process would get the list of device specific handlers, in that case.
As it is implemented now, `gokr-packer` is responsible for parsing the device manifest and passing along handler information to `gokrazy` via `main()` in the generated `init,go`.

There are a few other ways to achieve this:
1. Store relevant config in some kind of `map[deviceType]deviceDetails` in `gokrazy/gokrazy`. This way, we can get rid of the device manifest file altogether and pass along `mbr-but-not-gpt` and "device type" as flags to `gokr-packer`. Makes it easy to call `gokrazy.SetupDeviceSpecifics(deviceType)` via `init.go` in this approach. This is the simplest way to implement it. Sounds like this is what you're suggesting.
2. Keep the device manifest around, dump the manifest file in the root image and have it parsed and acted on by `gokrazy`. Instead of passing along arbitrary update handlers, `gokrazy` just reads the (optional) manifest on start and registers appropriate handlers.

What do you think?

stapelberg:
> Hmmm, I'm trying to figure out where gokrazy process would get the list of device specific handlers, in that case.

What does https://pkg.go.dev/github.com/gokrazy/gokrazy#Model return on the odroid? (You can see the detected model in the gokrazy web interface.)

If the odroid devices report their model correctly, that might be the easiest way.

Keeping a map of device specifics inside gokrazy (option 1 in your reply) sounds preferable over option 2 to me.

anupcshan:
> > Hmmm, I'm trying to figure out where gokrazy process would get the list of device specific handlers, in that case.
> 
> What does https://pkg.go.dev/github.com/gokrazy/gokrazy#Model return on the odroid? (You can see the detected model in the gokrazy web interface.)
> 
> If the odroid devices report their model correctly, that might be the easiest way.

They return "Hardkernel Odroid HC1" (from https://github.com/torvalds/linux/blob/c9e6606c7fe92b50a02ce51dda82586ebdf99b48/arch/arm/boot/dts/exynos5422-odroidhc1.dts#L14). That's unique enough to identify this device type.

> 
> Keeping a map of device specifics inside gokrazy (option 1 in your reply) sounds preferable over option 2 to me.

Yep, I've gone ahead and stored the device configs in `gokrazy/internal` - https://github.com/gokrazy/internal/pull/9. Its not in `gokrazy/gokrazy` because `gokr-packer` also needs to reference this device config while generating disk image/updating.

Let me make the change to use `gokrazy.Model` as the device type key (and rename that to device model).

anupcshan:
OK, PRs for all relevant changes are ready now (in dependency order):
1. https://github.com/gokrazy/internal/pull/9 has the device config itself with some validation checks
2. This PR calls `gokrazy.SetupDeviceSpecifics` in `SuperviseServices` - uses `gokrazy.Model` to automatically register device-specific stuff.
3. https://github.com/gokrazy/tools/pull/28 to do the gokr-packer bits.

Each PR depends on the one before it (needs a go mod upgrade to pick up previous change).

I use https://github.com/anupcshan/odroidbake/blob/main/Makefile to manage a test installation.

update.go
```diff
@@ -240,6 +286,11 @@ func initUpdate() error {
 	http.HandleFunc("/update/root", nonConcurrentUpdateHandler(rootdev.Partition(rootdev.InactiveRootPartition())))
 	http.HandleFunc("/update/switch", nonConcurrentSwitchHandler(rootdev.InactiveRootPartition()))
 	http.HandleFunc("/update/testboot", nonConcurrentTestbootHandler(rootdev.InactiveRootPartition()))
+
+	for _, extraUpdateHandler := range extraUpdateHandlers {
+		http.HandleFunc(fmt.Sprintf("/update/%s", extraUpdateHandler.name), nonConcurrentLimitedUpdateHandler(extraUpdateHandler.device, extraUpdateHandler.offset, extraUpdateHandler.maxLength))
```
stapelberg, line 291: I think an extra level of hierarchy would be good in the paths. How about `/update/device-specific/<name>`?

update.go
```diff
@@ -225,6 +243,34 @@ func nonConcurrentTestbootHandler(newRootPartition int) func(http.ResponseWriter
...
+var extraUpdateHandlers []*ExtraUpdateHandler
+
+func setupDeviceSpecifics() {
+	modelName := Model()
+	cfg, ok := deviceconfig.DeviceConfigs[modelName]
+	if !ok {
+		return
+	}
+
+	log.Printf("setting up device-specific handlers for '%s'", modelName)
```
stapelberg, line 262: Use %q instead of '%s' here

stapelberg, line 262: Make it “device-specific update handlers” instead of “device-specific handlers”

update.go
```diff
@@ -225,6 +243,34 @@ func nonConcurrentTestbootHandler(newRootPartition int) func(http.ResponseWriter
 	}
 }
 
+type ExtraUpdateHandler struct {
```
stapelberg, line 246: Why is this type exported? It doesn’t need to be, right? Can you change it to `extraUpdateHandler` (lower-case e)?

stapelberg (commented):
This is looking much better now!

stapelberg (approved):

stapelberg:
This PR is good to go! Can you remove the `replace` directive in go.mod and pull in the latest `gokrazy/internal` please?

anupcshan:
> This PR is good to go! Can you remove the `replace` directive in go.mod and pull in the latest `gokrazy/internal` please?

Thanks for the thorough review! Updated.

stapelberg (approved):

PR #105, opened 2022-01-06
supervise: added optional configuration to wait for clock

dtchanpura:
For those services that need accurate time to have proper details but cannot be changed easily.

~~Based on the environment variable (`GOKRAZY_WAIT_FOR_CLOCK`), that can be added at the time of configuration, the service will not be started unless the clock has been set properly (using `WaitForClock()` function).~~

Based on the configuration file under `waitforclock/<package>/waitforclock.txt` the service will not be started until the clock has been set properly (using `WaitForClock()`) function)

This problem was faced when `vmagent` was started up with the metrics value of `process_start_time_seconds 7` which when we do difference gives over 50+ years of uptime and is inaccurate.

Note: The same problem was not found when running `prometheus` because of the difference in underlying code.


stapelberg (changes requested):
I don’t mind making the wait-for-clock behavior configurable, but I’d rather not overload the environment variable mechanism for it. 

Another piece of config that is configurable is the “dontstart” setting of the packer. Can you implement the “waitforclock” setting like the “dontstart” one please?

dtchanpura:
Yeah, I do agree on overloading environments. I was not sure if this would've been the best way.

Let me try to add a `waitforclock` setting. I will follow up with questions, if any.

As far as I understand, it will need some changes in both gokrazy/tools repository and this one, correct?

stapelberg:
> As far as I understand, it will need some changes in both gokrazy/tools repository and this one, correct?

Yep!

Let me know if you get stuck.

dtchanpura:
@stapelberg I have implemented the change in both repositories and tested it on Raspberry Pi 4 with `dontstart` and then `waitforclock`, I do have some non-blocking questions:

1. Does the function name look good? `NewDelayedService`
2. Do we need to change service.waitForClock to false after we have waited for clock? -- I would feel it's fine, it will just log the message waiting for clock and not actually wait on the restart.

dtchanpura:
I am trying to test it but it seems to be going in all services stopped mode and have to reboot manually by switching it off and on. Will update here if it is a problem because of this or it is something else.

stapelberg:
> 1. Does the function name look good? `NewDelayedService`

Instead of focusing on the effect (delayed), I would prefer if the name focused on the intent (wait for clock). Can you make it NewWaitForClockService please?

> 2. Do we need to change service.waitForClock to false after we have waited for clock? -- I would feel it's fine, it will just log the message waiting for clock and not actually wait on the restart.

Agreed, no need to change it, as subsequent waits will be instant.

dtchanpura:
@stapelberg Added the changes as per comments, also synced the master code to resolve conflicts.

stapelberg (approved):

PR #106, opened 2022-01-12
supervise: wait for services to stop before reboot

anupcshan:
Currently, rebooting/upgrading the instance sends a TERM signal to all
running services, waits 1 second and then reboots. Processes that take
longer to cleanly stop can't do so.

Instead, give each service up to 15s to do a clean shutdown. Reboot 15s
after the command is sent, or after all services are stopped, whichever
comes first.

To make it easier to kill entire process tree for a service, create a
per-service process group. Stopping the service can be done by sending a
single signal to the process group.

### DIFF
--- a/proc.go
+++ b/proc.go
@@ -1,0 +1,48 @@
+package gokrazy
+
+import "sync"
+
+type statusCode int32
+
+const (
+	Stopped  statusCode = iota // Process not running
+	Running                    // Process was started and is very likely still running
+	Stopping                   // Process is being stopped, but it might still be running
+)
+
+type processState struct {
+	lock         sync.Mutex
+	statusChange *sync.Cond
+	status       statusCode
+}
+
+func NewProcessState() *processState {
+	p := &processState{}
+	p.statusChange = sync.NewCond(&p.lock)
+
+	return p
+}
+
+func (p *processState) Set(status statusCode) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	if p.status == Stopped && status == Stopping {
+		// Not a valid state transition. Ignore it.
+		return
+	}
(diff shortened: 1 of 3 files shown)


anupcshan:
Apart from waiting for services to stop before reboot, creating service processes in a process group is an important change happening here. It is required to make sure `cmd/ntp` exit cleanly - this process cannot be restarted in the UI right now. With this change, it can. The current behavior only sends a `TERM` signal to the top-level process, which doesn't get propagated to its children. If the top-level process is waiting for children to exit, we either need to `KILL` it or send a signal to its children.

A downside of using a process group and sending a signal to the entire PG is that parent process can no longer "protect" its children. But, I doubt services running under gokrazy will care about this.

Let me know if this makes sense.

stapelberg (approved):

stapelberg:
Thanks! I notice that `ntp` isn’t cleanly stopped, though: another `[ntp]` process stays behind each time I stop `ntp`:

```
  910 0         2140 S    {exe} sh
 1955 65534        0 Z    [ntp]
 2035 0         686m S    /gokrazy/ntp
 2040 65534     687m S    /gokrazy/ntp
 2051 0         2140 R    {exe} ps

```

anupcshan:
> Thanks! I notice that `ntp` isn’t cleanly stopped, though: another `[ntp]` process stays behind each time I stop `ntp`:

Oh, interesting. I *think* this might need a `waitpid(-pgid)` (`syscall.Wait4`?) instead of `waitpid(pid)` (which is what `os.Process.Wait()` is doing). Otherwise, it might be possible for child processes to wait to be cleaned up.

I'll take a look in a few hours and send a fix.

PR #107, opened 2022-01-13
supervise: wait for child processes

anupcshan:
With gokrazy/gokrazy#106, we're sending a TERM signal to service
processes and their immediate children. Service processes are being
waited on by `cmd.Wait()`, which will reap any kernel state associated
with it.  However, its child processes don't necessarily get reaped
(depending on whether its parent had time to clean up before exiting).

With this change, we explicitly wait for everything in the service
process group to exit and prevent zombies.

### DIFF
--- a/supervise.go
+++ b/supervise.go
@@ -427,8 +427,9 @@ func supervise(s *service) {
 		s.state.Set(Running)
 		s.setProcess(cmd.Process)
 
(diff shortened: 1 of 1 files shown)


stapelberg (approved):

stapelberg:
Thanks!

PR #108, opened 2022-01-27
integration: add test for MBR-only booting

anupcshan:
Refactor integration/uefiboot to be reusable.

Related to https://github.com/gokrazy/gokrazy/issues/101#issuecomment-1012831153


integration/common/testutils.go
```diff
@@ -0,0 +1,139 @@
+package common
```
stapelberg, line 1: Per https://github.com/golang/go/wiki/CodeReviewComments#package-names, avoid package names like `common`.

I suggest `integrationtest` as package name, and `integrationtest.Run` as function name.

go.mod
```diff
@@ -16,3 +16,6 @@ require (
 	golang.org/x/sys v0.0.0-20210820121016-41cdb8703e55
 	rsc.io/goversion v1.2.0
 )
+
+// TODO: Remove after https://github.com/gokrazy/internal/pull/11 is merged
+replace github.com/gokrazy/internal => github.com/anupcshan/internal v0.0.0-20220127050546-b923a796f0ca
```
stapelberg, line 21: This can now be updated please :)

stapelberg (changes requested):

stapelberg (approved):

PR #112, opened 2022-03-03
Docs: configure wifi via extrafile

christian-heusel:
This pullrequest simplifies the wifi setup by using the extrafiles instead of mounting the `/perm`-partition.


stapelberg (approved):

stapelberg:
Thanks!

PR #115, opened 2022-03-09
website: update Tailscale VPN documentation

joneskoo:
Replace use of breakglass with instructions to use interactive login.
This depends on tailscale/tailscale#4117 being merged.

Add mkfs into the example so that /perm filesystem is initialized.

Add instructions how to use auth-key for unattended device auth instead
of having to use browser interactively.

Append to tailscaled/flags.txt instead of overwriting.

Clarify and update the tsnet example. Add mention of TS_AUTHKEY
as alternative here too.

### DIFF
--- a/website/content/userguide/install/tailscale.md
+++ b/website/content/userguide/install/tailscale.md
@@ -19,66 +19,39 @@ remotely (no static DHCP leases, port-forwarding and DynDNS required!), or even
 to secure your communication when gokrazy is [connected to an unencrypted WiFi
 network](/userguide/unencrypted-wifi/).
 
-## Step 1. set command-line flags
-
-We need to specify the following flags for the `tailscaled` daemon (see [Package
-config: flags and environment variables](/userguide/package-config) if you’re
-unfamiliar with this mechanism):
-
-```shell
-mkdir -p flags/tailscale.com/cmd/tailscaled
-cat > flags/tailscale.com/cmd/tailscaled/flags.txt <<EOF
---statedir=/perm/tailscaled/
---tun=userspace-networking
-EOF
-
-mkdir -p flags/tailscale.com/cmd/tailscale
-echo 'up' > flags/tailscale.com/cmd/tailscale/flags.txt
-```
-
-Tailscale requires a writable directory used by `tailscaled` to store its state.
-It is used to persist authentication over reboots and for control socket
-`tailscaled.sock`.
-
-The `--tun=userspace-networking` flag selects the [Userspace
-networking](https://tailscale.com/kb/1112/userspace-networking/) mode because
+{{% notice note %}}
+Tailscale currently uses [Userspace networking] mode on gokrazy, because
 for `tun` mode, Tailscale currently requires components that gokrazy does not
 provide. For accessing the services on your gokrazy installation, the Userspace
 networking mode works fine, though 🥳 .
 
-## Step 2. include the tailscale packages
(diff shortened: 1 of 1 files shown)


website/content/userguide/install/tailscale.md
```diff
@@ -53,25 +55,47 @@ have one yet), include the Tailscale daemon and CLI Go packages:
...
+  github.com/gokrazy/mkfs \
   tailscale.com/cmd/tailscaled \
   tailscale.com/cmd/tailscale
 ```
 
-## Step 3. authenticate
+We include mkfs to create filesystem on the /perm partition. mkfs only needs
+to be done once.
+
+## Step 3a. authenticate interactively
```
stapelberg, line 66: Instead of naming these Step 3a and Step 3b (which could be misinterpreted as having to do both), can we make it more clear that one of the two options is enough? Maybe “Step 3 Alternative A” and “Step 3 Alternative B”?

stapelberg (commented):
Thanks!

website/content/userguide/install/tailscale.md
joneskoo: I updated based on bradfitz comment re: tailscale v1.22.1 and also edited the steps.

Hopefully this version is better.

website/content/userguide/install/tailscale.md
```diff
@@ -19,58 +19,88 @@ remotely (no static DHCP leases, port-forwarding and DynDNS required!), or even
...
   github.com/gokrazy/hello \
-  github.com/gokrazy/breakglass \
-  github.com/gokrazy/serial-busybox \
+  github.com/gokrazy/mkfs \
   tailscale.com/cmd/tailscaled \
   tailscale.com/cmd/tailscale
 ```
 
-## Step 3. authenticate
+We include `mkfs` to automatically initialize filesystem on the `/perm`
```
stapelberg, line 80: …initialize **a** file system…

website/content/userguide/install/tailscale.md
```diff
@@ -19,58 +19,88 @@ remotely (no static DHCP leases, port-forwarding and DynDNS required!), or even
...
 
-`tailscaled` requires the `--statedir` flag, so we need to set it
-explicitly. `/perm/tailscaled` is the working directory of the `tailscaled`
-process and will contain the `tailscaled.sock` socket, so it makes sense to
-place the state file into the same directory.
+**Option B: unattended authentication with auth key**
 
-The `--tun=userspace-networking` flag selects the [Userspace
-networking](https://tailscale.com/kb/1112/userspace-networking/) mode.
+Navigate to [Tailscale console] and open Settings / Keys. Generate auth key.
```
stapelberg, line 56: The `[Tailscale console]` link seems broken when previewing the website locally.

stapelberg, line 56: Can we prefix this paragraph with “Alternatively, ” please

stapelberg (changes requested):
Thanks, just a few small adjustments :)

website/content/userguide/install/tailscale.md
joneskoo: A-ha. It worked when I wrote it but broke when I moved the other section with same link to notice box. Apparently that scope is somehow different. Good catch.

joneskoo:
Fixed review comments.

stapelberg:
Now live, thanks!

Issue #49, opened 2019-08-25
Bluetooth support


joneskoo:
## Platform

I’m using:

* [ ] gokrazy/rpi3b
* [X] gokrazy/rpi3b+
* [X] gokrazy/rpi02w
* [ ] gokrazy/apu2c4
* [ ] gokrazy/x86-64

## Observed behavior

I installed github.com/joneskoo/ruuvi-prometheus  which depends on Bluetooth hardware. It failed to start.

After breaking glass, and looking at the kernel, it looks like Bluetooth is intentionally disabled.

I could not find an option to disable serial console and enable Bluetooth instead. I know they can't be enabled at the same time on the hardware UART, possibly not at all.

On Alpine Linux, I know I need to use bluez btattach to initialize the Bluetooth interface on serial port. Online, I saw pointers that it could be enabled without bluez using dts overlay: https://github.com/RoEdAl/alarm-bluetooth-raspberrypi

## Expected behavior

Raspberry Pi 3B+ Bluetooth supported (as option)

stapelberg:
> After breaking glass, and looking at the kernel, it looks like Bluetooth is intentionally disabled.

No, it just never was intentionally enabled, and we started from a minimal kernel config :)

I don’t use bluetooth much, so I’m not sure what needs to be done to make this work.

Maybe you could try a few things and share how it’s going? Sounds like you have a few good pointers already (dts changes, freeing up the serial port).

joneskoo:
I assume it will be a requirement to keep the serial port console? So
Bluetooth implicitly needs to be behind some flag?

If I find time to hack it to enable it as a proof of concept, would you
consider including such an option? I’m not sure if/when I’ll find such time
but will not do it if we know it won’t be included as option.
-- 
Joonas Kuorilehto

stapelberg:
> I assume it will be a requirement to keep the serial port console?

Yes.

> So Bluetooth implicitly needs to be behind some flag?

Possibly. Our serial console setup is a bit different from the Raspberry Pi’s default setup, so I can’t say which changes are required.

> If I find time to hack it to enable it as a proof of concept, would you consider including such an option?

I will definitely consider it. Bluetooth support would be great to have.

joneskoo:
Sharing here we discussed this on Twitter: https://twitter.com/joneskoo_yx/status/1500152012536795136?s=21

What I tried so far:

```diff
[...]

PR #116, opened 2022-03-09
website: add Bluetooth to user guide

joneskoo:
Related to #49 .


website/content/userguide/bluetooth.md
```diff
@@ -0,0 +1,56 @@
...
+```shell
+gokr-packer \
+  -tls=self-signed \
+  -update=yes \
+  github.com/gokrazy/hello \
+  github.com/gokrazy/bluetooth
+```
+
+The `github.com/gokrazy/bluetooth` package loads the appropriate kernel modules
+and firmware required.
```
joneskoo, line 26: Technically there's no firmware required… But the package should load firmware if it was required/beneficial. That it doesn't now, maybe technical detail and it's ok to say here it loads the firmware required, which currently is nothing?

stapelberg: Filed https://github.com/gokrazy/bluetooth/issues/2 so that we don’t forget. Yeah, I think it’s okay to state it here already.

stapelberg (approved):

PR #120, opened 2022-03-24
Improve authenticated

oliverpool:
_I want to contribute to this project and I thought that I could get started with a minor PR. Feel free to tell me what I could do better_

I took a look at `authenticated.go` and found 3 minors improvements that could be made:
1. always send the `WWW-Authenticate` on authentication failure. Without this, firefox is stuck on the 401 page and does not prompt the user for a new password (it sends the same password every time). With this change it instead asks the user for a new password every time. If the user clicks "cancel", then the error message is displayed.
2. use the newly added [`strings.Cut`](https://pkg.go.dev/strings#Cut) (since go.mod already indicates `go 1.18` I guess this is ok)
3. compare the password in constant time (to mitigate timing attack)

Since those 3 changes are pretty small, I gathered them in one PR (but separated commits). If you prefer, I can split the commits into multiple PRs.

### DIFF
--- a/authenticated.go
+++ b/authenticated.go
@@ -22,6 +22,7 @@ func authenticated(w http.ResponseWriter, r *http.Request) {
 
 	b, err := base64.StdEncoding.DecodeString(s[1])
 	if err != nil {
+		w.Header().Set("WWW-Authenticate", `Basic realm="gokrazy"`)
 		http.Error(w, fmt.Sprintf("could not decode Authorization header as base64: %v", err), http.StatusUnauthorized)
 		return
 	}
@@ -30,6 +31,7 @@ func authenticated(w http.ResponseWriter, r *http.Request) {
 	if len(pair) != 2 ||
 		pair[0] != "gokrazy" ||
 		pair[1] != httpPassword {
+		w.Header().Set("WWW-Authenticate", `Basic realm="gokrazy"`)
 		http.Error(w, "invalid username/password", http.StatusUnauthorized)
 		return
 	}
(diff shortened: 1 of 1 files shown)


authenticated.go
```diff
@@ -13,23 +14,25 @@ func authenticated(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "httpPassword not set", http.StatusInternalServerError)
 		return
 	}
-	s := strings.SplitN(r.Header.Get("Authorization"), " ", 2)
-	if len(s) != 2 || s[0] != "Basic" {
+	kind, encoded, found := strings.Cut(r.Header.Get("Authorization"), " ")
```
stapelberg, line 17: Can you leave this as-is for now and leave a comment to convert it to strings.Cut instead when Go 1.19 is out?

We generally support the current stable release and (best-effort) the previous stable release, which is Go 1.17 in this case.

stapelberg (changes requested):
Thanks for your PR!

authenticated.go
oliverpool: Sure, I will revert this.

If you want to support go 1.17, you should change the following line in `go.mod`:
https://github.com/gokrazy/gokrazy/blob/4c1cf82b51a18e58718dd34de5d56d3c205b8547/go.mod#L3

https://go.dev/doc/modules/gomod-ref
> The minimum version of Go required by the current module.

stapelberg: Can you change it while you’re touching this?

oliverpool: Done (I simply copied the original function in the meantime ;)

stapelberg (approved):

oliverpool:
Thank you for the very quick review!

PR #121, opened 2022-03-26
website: document community supported platforms

oliverpool:
_Closes #119_ 

I was not sure how to integrate this into the current "supported platforms" page, so I ended up creating a new category "Community-supported platforms".

Let me know what you think!

### DIFF
--- a/website/content/platforms/_index.markdown
+++ b/website/content/platforms/_index.markdown
@@ -63,6 +63,35 @@ The leading <code>github.com</code> in front of package import paths has been om
 [stapelberg/hmgo]: https://github.com/stapelberg/hmgo
 [stapelberg/scan2drive]: https://github.com/stapelberg/scan2drive
 
+## Community-supported platforms
+
+To use an alternative kernel (or firmware), adjust the `-kernel_package` flag (or `-firmware_package` flag) of the `gokr-packer` tool, to point to the package:
+
+```
+GOARCH=arm gokr-packer \
+    -kernel_package=github.com/gokrazy-community/kernel-rpi-os-32/dist \
+    -firmware_package=github.com/gokrazy-community/firmware-rpi/dist \
+    github.com/gokrazy/hello
+```
+
+Please report any issue to their respective repostitory.
+
+### Alternative kernels
+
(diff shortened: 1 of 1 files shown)


website/content/platforms/_index.markdown
```diff
@@ -63,6 +63,35 @@ The leading <code>github.com</code> in front of package import paths has been om
 [stapelberg/hmgo]: https://github.com/stapelberg/hmgo
 [stapelberg/scan2drive]: https://github.com/stapelberg/scan2drive
 
+## Community-supported platforms
+
+To use an alternative kernel (or firmware), adjust the `-kernel_package` flag (or `-firmware_package` flag) of the `gokr-packer` tool, to point to the package:
```
stapelberg, line 68: Instead of listing all these details here, I think it’d be better to just link to a README file of each community-supported platform (the odroid port is another one I’m aware of, by the way).

Also, there are some questions that many readers will have, such as “under which circumstances would I use this over the standard gokrazy repositories” or “why is this not part of gokrazy itself”, and I think it’d be good to answer these questions here :)

I can write up a paragraph about that, or you can make a suggestion — either way is fine with me.

stapelberg (changes requested):
Thanks for your PR!

website/content/platforms/_index.markdown
oliverpool: > just link to a README file

You mean that I should simply delete the `Architecture` column?
I think having some kind of details here is nice, because it let the reader now, if there is any interest at all following the link to the readme (but maybe the "Supported platforms" column is enough).

I reworked the example, not to include any specific repository (`...` instead), to show how easy it is to use alternative kernel/firmware.

> there are some questions

I replaced the "Release policy" column with a "Raison d'être" (I'm not very happy with this title yet...), to explain why this alternative exists and how it is different from the official kernel/firmware.

> I can write up a paragraph about that

Sure, feel free to take the PR from here, and add/remove/change whatever you like :)

stapelberg:
Thanks, I have picked up this commit and modified it a little to — I hope — better explain when to use the community-supported kernel. Please send corrections if needed :) https://gokrazy.org/platforms/#community

PR #122, opened 2022-03-27
Run go mod tidy to fill in missing go.sum entries

hugelgupf:
For some reason, Go 1.16 is stricter with go.sum entries than Go 1.15 or Go 1.17:

```
$ go1.16 list -json -- ./cmd/dhcp
go: github.com/anatol/vmtest@v0.0.0-20211215032353-afd7b1dd38ef requires
        github.com/stretchr/testify@v1.7.0: missing go.sum entry; to add it:
        go mod download github.com/stretchr/testify
```

go1.15 and go1.17 both work. Honestly - don't ask me why, but this fixes it.


stapelberg:
Will using `go mod tidy` with Go 1.18 revert this change?

In general, we only support stable and previous stable, i.e. Go 1.18 and Go 1.17 currently.

hugelgupf:
I ran the go mod tidy using Go 1.18.

stapelberg:
Thanks!

PR #127, opened 2022-04-05
supervisor: stop service on exit 0

### LINKED ISSUE #125: Supervisor restart policy
## Platform

* [x] gokrazy/rpi1b

## Observed behavior

Services are always restarted, until they [exit with code 125](https://gokrazy.org/userguide/process-interface/).

## Expected behavior

Services are always restarted, until they exit with code 0 (no error).

---

I'm interested to know what is the rationale behind the current behavior (always restart, except on exit 125). I would rather have expected the gokrazy supervisor to restart failing processes only.

For instance in tailscale there is a specific case handling for gokrazy, which doesn't currently work on my slow rapsberry pi 1st gen. The first start of the process fails because tailscaled is not ready yet and never gets restarted:  https://github.com/tailscale/tailscale/pull/4117#issuecomment-1086641950

This could be fixed in tailscale, but in this case I think it would be nicer to change the behavior (or at least make it configurable) of gokrazy to restart a process until it succeeds. This would mean that the custom code for gokrazy could be removed from tailscale.

---

Since making this the default would be a breaking change, maybe this could be make configurable?
Either a the platform level (flag for packer), or at the service level (along the lines of https://gokrazy.org/userguide/package-config/)

I currently think of the following policies:
- restart on failure (the behavior I need - for all my programs, currently)
- stop on 125 (current behavior)
- always restart

Do you think this is a valid concern? If yes, how do you think should the be tackled?

oliverpool:
fixes #125

> I think we actually can do the dont-supervise-on-exit-0 change. It’s not really a case I typically have with the programs I run on gokrazy.

> For context, the platform was built for services that run indefinitely, with exit code 125 introduced initially for breakglass which should be off-by-default.

### DIFF
--- a/supervise.go
+++ b/supervise.go
@@ -473,7 +473,8 @@ func supervise(s *service) {
 			}
 			l.Println("gokrazy: " + err.Error())
 		} else {
-			l.Printf("gokrazy: exited successfully")
+			l.Printf("gokrazy: exited successfully, stopping")
+			s.setStopped(true)
 		}
 
 		if s.supervisionMode() == superviseOnce {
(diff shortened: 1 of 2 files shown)


stapelberg:
Thanks!

PR #128, opened 2022-04-06
website: category reorganization

### LINKED ISSUE #126: website: category reorganization suggestion
I find it great that so much is documented on the [gokrazy website](https://gokrazy.org)

However I find the current section ordering a bit counter-intuitive:

```
1. Showcase
2. Quickstart
3. Userguide
  - input/output
  - package config
  - process interface
  - remote syslog
  - TLS
  - wifi
  - bluetooth
  - go modules
  - Access permanent data
  - How do I install X on gokrazy (tailscale / caddy / Min.io)
4. Hardware and CI
5. Non-Go Prototyping
```

The resources to install an existing package (`how do I install X`), to configure gokrazy (`package config`, `wifi`) and to develop a new package (`process interface`, `bluethooth`, `Non-Go Prototyping`) are interleaved.

---

I would suggest to reorganize the categories, based on the usage : available packages / packaging existing / developing a new package.

For instance:
```
1. Quickstart (unchanged)
2. Supported platforms (Hardware and CI)
3. Configuration
  - package config (unchanged)
  - remote syslog (unchanged)
  - TLS (unchanged)
  - wifi (unchanged)
  - Access permanent data (unchanged)
4. Available packages (Showcase)
  - tailscale (unchanged)
  - caddy (unchanged)
  - min.io (unchanged)
5. Packaging new software (Non-Go Prototyping)
5. Package development
  - input/output (unchanged)
  - process interface (unchanged)
  - bluetooth (unchanged)
  - go modules (unchanged)
 ```

What do you think? I could make a PR, if you think that this would be an improvement (can hugo handle redirects, to prevent 404s?).

### PR DESCRIPTION
**Pull request by oliverpool**
fixes #126

This is what I currently have:
```
1. Quickstart (unchanged)
2. Supported platforms (Hardware and CI)
3. Configuration
  - package config (unchanged)
  - remote syslog (unchanged)
  - TLS (unchanged)
  - wifi (unchanged)
  - bluetooth (unchanged)
  - Access permanent data (unchanged)
4. Available packages
  - Showcase (unchanged)
  - tailscale (unchanged)
  - caddy (unchanged)
  - min.io (unchanged)
5. Package development
  - process interface (unchanged)
  - input/output (unchanged)
  - go modules (unchanged)
  - Non-Go Prototyping (unchanged)
```

Compared to my initial suggestion, here are the adjustments:
- "Available packages" is a new page, with a short explanation on how to add go commands
- removal of the category "Packaging new software" (content is in "Available packages")
- “Showcase” a sub page under “Available packages” (as you suggested)
- "Non-Go Prototyping" is a sub-page of "Package development" (name as you suggested, a bit more hidden as a sub-page, since it’s "supposed to be a stop-gap")
- bluetooth is under "configuration" (as you suggested)

I did some content changes, but tried to limit them as much as possible, to tackle first the category structure.

Once you are happy with it, we can see if some text need to be updated.

### CODE REVIEW

**Comment by oliverpool** [comment]
I just realized, that my code editor (vscode) did some auto-formatting on the markdown content... Should I revert them?

File: website/content/packages/_index.markdown [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+---
+
+# Available packages
+
+The guides in this chapter show how to set up popular third-party software.
+
+If you feel a common program should be included here but isn’t, please [file an
+issue on GitHub](https://github.com/gokrazy/gokrazy/issues).
+
+## Installing existing Go command
```
stapelberg, line 15: Go command → Go software

website/content/prototyping/_index.markdown
```diff
@@ -1,195 +1,17 @@
...
-software written in Go. If there is no Go software for what you want
-to do, creating that piece of software can pose a seemingly
-unsurmountable hurdle. To make some quick progress and figure out if
-your idea can be implemented, it might make sense to temporarily use
-existing software before starting your own implementation.
+# Package development
 
-This article shows a couple of techniques for getting non-Go
-software to work on gokrazy, in increasing order of complexity.
+A gokrazy package is simply the `main` package of a Go software.
```
stapelberg, line 11: That’s not entirely correct, actually. A gokrazy package is a Go package.

Even if the Go package declares a named library instead of a `main` package, that’s still a valid gokrazy package, and can be useful to install. https://github.com/gokrazy/podman is an example for that: it declares `package podman` in `podman.go`, so it won’t result in an extra binary (no `/user/podman`). But, because it also contains a `_gokrazy/extrafiles.tar`, those files will be included in the root file system.

website/content/prototyping/_index.markdown
```diff
@@ -1,195 +1,17 @@
...
-here **will not be automatically updated** by gokrazy
-and hence poses a security risk. Use these techniques only for
-prototyping.
-	  
-## Go software not written for gokrazy: node-exporter
-	  
-The Prometheus node-exporter doesn’t use cgo and needs no
-command-line parameters, configuration files or other assets can be
-added to the <code>gokr-packer</code> command line.
+To use the local version of a repository, simply use the [replace directive](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) of `go.mod`.
```
stapelberg, line 13: Here and above, remove “simply” — it’s subjective and doesn’t add anything :)

website/content/userguide/_index.markdown
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Userguide"
+title: "Configuration"
```
stapelberg, line 2: The rest of this file still mentions “Userguide”. That should probably be updated to “Configuration” as well.

stapelberg (changes requested):
Thanks for this PR!

Can you add `aliases:` directives to all files that you moved? That way, the links should keep working.

I don’t mind the markdown formatting changes, so no need to revert them.

oliverpool:
Thanks for the review!

I have added `aliases` to all the page that were (re)moved, I think I didn't miss any.

I also replaced `userguide` with `configuration guides`.

website/content/prototyping/_index.markdown
```diff
@@ -1,195 +1,17 @@
...
-software written in Go. If there is no Go software for what you want
-to do, creating that piece of software can pose a seemingly
-unsurmountable hurdle. To make some quick progress and figure out if
-your idea can be implemented, it might make sense to temporarily use
-existing software before starting your own implementation.
+# Package development
 
-This article shows a couple of techniques for getting non-Go
-software to work on gokrazy, in increasing order of complexity.
+A gokrazy package is simply a package of a Go software.
```
stapelberg, line 11: We can make this a bit more tangible. How about:

A gokrazy package corresponds to a Go module. The `gokr-packer` program is a wrapper around `go install`.

stapelberg (changes requested):
Thanks, almost there :)

website/content/prototyping/_index.markdown
```diff
@@ -1,195 +1,17 @@
 ---
 layout: default
-title: Non-Go Prototyping
+title: Package development
```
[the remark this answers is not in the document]
oliverpool: I kept the `/prototyping/` URI, since the current `Non-Go Prototyping` page actually starts with go sections:


The `Go software not written for gokrazy:` remains available at this URI (only the non-go stuff changed location).

I am not found of using `/development/` for this. I think it would be better to document how to contribute to gokrazy under this URI (supervisor, packer...).

What do you think?

website/content/prototyping/_index.markdown
oliverpool: > A gokrazy package corresponds to a Go module. 

I think a Go module can have multiple packages, so this is also not optimal.

For instance the `github.com/VictoriaMetrics/VictoriaMetrics` module as many packages: https://github.com/VictoriaMetrics/VictoriaMetrics/tree/master/app

stapelberg: Right, let’s make it:

We can make this a bit more tangible. How about:

A gokrazy package corresponds to a Go package (which in turn is part of a Go module). The `gokr-packer` program is a wrapper around `go install`.

website/content/prototyping/_index.markdown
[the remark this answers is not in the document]
stapelberg: I think `/contributing/` would be better for documentation about contributing :) In general, the URL should match the title.

I understand the split between Go and non-Go software, but I still think the `/prototyping/` URL should lead to the non-go prototyping instructions — I have linked to it in various contexts, so it would not be great if we moved the content away.

oliverpool:
Back from holidays, I addressed your 2 comments.

I did a squash and a rebase on master (since you added the rsync-backup page in the meantime).

stapelberg:
Thanks very much! Now merged, should be live within a few minutes

PR #134, opened 2022-06-01
Add krazyotelcol to "Written for gokrazy"

svrnm:
Stumbled upon gokrazy the other day watching @stapelberg's video on "Why I wrote my own rsync", I immediately wanted to give it a try with the OpenTelemetry collector. *krazyotelcol* is the result of that!

Thanks for this amazing project :-)


svrnm:
Awesome, thanks! Next I need to put prometheus, otelcollector & grafana together ;-)

Issue #145, opened 2022-09-22
Userguide: Search isn't working


mhofstetter:
## Observed behavior

The search via the field in the side navigation isn't working.

## Expected behavior

I'd expect that the search works - or otherwise is disabled in the theme.

This helps to find the information about all the awesome features of gokrazy :) Keep up the good work!

stapelberg:
Thanks! Now merged

PR #146, opened 2022-09-22
userguide: enable search by generating index.json as output

mhofstetter:
This should enable the search in the GoKrazy userguide and solves https://github.com/gokrazy/gokrazy/issues/145.

See https://learn.netlify.app/en/basics/configuration/#activate-search for further information.

I guess that you're going to re-generate the documentation as you did with in the past - which then should generate the `index.json`.


stapelberg (approved):

stapelberg:
Thanks!

Issue #164, opened 2023-01-26
Support passing flags to `gok run`


markdrayton:
## Platform

I’m using:

* [ ] gokrazy/rpi3b
* [ ] gokrazy/rpi3b+
* [x] gokrazy/rpi4b
* [ ] gokrazy/apu2c4
* [ ] gokrazy/x86-64

## Observed behavior

`gok run` doesn't support passing flags to the binary it runs.

## Expected behavior

`cd ~/go/src/github.com/stapelberg/scan2drive/cmd/scan2drive && gok run -autocert_host_list=scan2drive.example.com` would run `scan2drive -autocert_host_list=scan2drive.example.com` on the target instance.

I had a quick look at how to do this but adding it to the current diversion approach doesn't seem straightforward. I suspect support for starting a new service (as mentioned at https://github.com/gokrazy/gokrazy/issues/88#issuecomment-974845682) is needed. I could potentially take a look at this if you could provide some high-level tips on where to wire things up.

stapelberg:
Hey, thanks for having a look at this improvement :)

With the instance-centric config now having landed, this should be feasible: `gok run` should take the current flags from the instance config, append the flags specified on the command line (if any), and include the entire list of flags in the divert HTTP request.

Because sending a slice of strings via HTTP form parameters is awkward, it might be easiest to define a DivertRequest struct that is then decoded/encoded using JSON, and send that as the divert request body.

The divert handler itself would need to be modified to either include the flags on the `cmd` when calling `NewService` (in case of a program that has not been included in the gokrazy instance yet), or call `setCmd` (which doesn’t exist yet) when diverting an existing service.

In terms of syntax, it would not be `gok run -autocert_host_list=scan2drive.example.com`, but `gok run -- -autocert_host_list=scan2drive.example.com` so that the flags are not confused with flags for `gok run` itself.

Does that roughly make sense? Let me know if you have detail questions.

markdrayton:
This sounds pretty reasonable, thanks! I'll take a look when I get a moment.

> In terms of syntax, it would not be gok run -autocert_host_list=scan2drive.example.com, but gok run -- -autocert_host_list=scan2drive.example.com so that the flags are not confused with flags for gok run itself.

Ah, yes, I was doing this in testing but forgot to report it accurately when I opened the issue. I assume cobra will handle this but I'll check either way.

markdrayton:
[...]

PR #166, opened 2023-02-01
Support passing command line flags via `gok run`

markdrayton:
See https://github.com/gokrazy/gokrazy/issues/164.

Still to do/decide:

1) What locking is needed for `s.cmd.Args`? I'm tempted to just wrap a mutex around it (or `s.cmd`) but I haven't satisfied myself that I fully understand all the code and implications. I'd take a suggestion here.

2) I fiddled with making `divertRequest` public but that led to `updater` depending on `gokrazy`, which it doesn't yet. Same deal in the opposite direction if I declare it in the `updater` package. Should I put it in the `internal` repo or is what I have fine?

3) I'm assuming we don't need a fully backwards- and forwards-compatible migration path for the changes in each of the 3 affected repos and can just commit (I think) the `gokrazy` and `updater` changes first, then update `go.mod` in `tools` to point at the updated dependencies.

Test run:

```sh
# after pointing github.com/gokrazy/{gokrazy,updater} at my local repos and building an updated `gok`
$ gok -i test update
[..]

$ cd ~/go/src/github.com/gokrazy; cat hello.go
package main

import (
	"log"
	"os"
)

func main() {
	log.Printf("%+v", os.Args)
}

$ jq '.PackageConfig."github.com/gokrazy/hello"' < ~/gokrazy/test/config.json
{
  "CommandLineFlags": [
    "boop"
  ]
}

$ gok -i test run -- a b c
[done] in 0.64s
Transferred hello (2 MiB) at 4.57 MiB/s (total: 0s)
2023/02/01 23:55:26 streaming logs of service "hello" from gokrazy instance "test"
2023/02/01 23:55:26 gokrazy: attempt 0, starting ["/tmp/gok-run/hello" "boop" "a" "b" "c"]
2023/02/01 23:55:26 [/tmp/gok-run/hello boop a b c]
2023/02/01 23:55:26 gokrazy: exited successfully, stopping
^C

$ gok -i test run -- a b c d
[done] in 0.48s
Transferred hello (2 MiB) at 4.57 MiB/s (total: 0s)
2023/02/01 23:55:30 streaming logs of service "hello" from gokrazy instance "test"
2023/02/01 23:55:26 gokrazy: attempt 0, starting ["/tmp/gok-run/hello" "boop" "a" "b" "c"]
2023/02/01 23:55:26 [/tmp/gok-run/hello boop a b c]
2023/02/01 23:55:26 gokrazy: exited successfully, stopping
2023/02/01 23:55:31 gokrazy: attempt 1, starting ["/tmp/gok-run/hello" "boop" "a" "b" "c" "d"]
2023/02/01 23:55:31 [/tmp/gok-run/hello boop a b c d]
2023/02/01 23:55:31 gokrazy: exited successfully, stopping
```

image

### DIFF
--- a/update.go
+++ b/update.go
@@ -394,23 +394,44 @@ func uploadTemp(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+type divertRequest struct {
+	Path      string   `json:"Path"`
+	Diversion string   `json:"Diversion"`
+	Flags     []string `json:"Flags"`
+}
+
 func divert(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodPost {
 		http.Error(w, "expected a POST request", http.StatusBadRequest)
 		return
 	}
 
-	diversion := r.FormValue("diversion")
-	if diversion == "" {
+	var req divertRequest
+	if r.Header.Get("Content-Type") == "application/json" {
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			http.Error(w, "couldn't read request body", http.StatusBadRequest)
+			return
+		}
+		err = json.Unmarshal(body, &req)
+		if err != nil {
+			http.Error(w, "couldn't unmarshal request body", http.StatusBadRequest)
+			return
+		}
+	} else {
+		req.Diversion = r.FormValue("diversion")
+		req.Path = r.FormValue("path")
+	}
+
+	if req.Diversion == "" {
 		http.Error(w, "diversion parameter is not set", http.StatusBadRequest)
 		return
 	}
 
-	path := r.FormValue("path")
-	svc := findSvc(path)
+	svc := findSvc(req.Path)
 	if svc == nil {
 		log.Printf("adding new service in-memory to make diversion work")
-		cmd := exec.Command(path)
+		cmd := exec.Command(req.Path, req.Flags...)
 		svc = NewService(cmd).s
 		services.Lock()
 		services.S = append(services.S, svc)
@@ -420,13 +441,16 @@ func divert(w http.ResponseWriter, r *http.Request) {
 
 	// Ensure diversion binary is executable (/uploadtemp creates regular,
 	// non-executable files).
-	diversion = filepath.Join(os.TempDir(), diversion)
+	diversion := filepath.Join(os.TempDir(), req.Diversion)
 	if err := os.Chmod(diversion, 0755); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
 	svc.setDiversion(diversion)
+	cmd := svc.Cmd()
+	cmd.Args = append([]string{cmd.Args[0]}, req.Flags...)
+	svc.setCmd(cmd)
 
 	if err := restart(svc, syscall.SIGTERM); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


stapelberg:
Thanks for your PRs!

> 1. What locking is needed for `s.cmd.Args`? I'm tempted to just wrap a mutex around it (or `s.cmd`) but I haven't satisfied myself that I fully understand all the code and implications. I'd take a suggestion here.

Now that we modify it, we need a mutex around `s.cmd`.

> 2. I fiddled with making `divertRequest` public but that led to `updater` depending on `gokrazy`, which it doesn't yet. Same deal in the opposite direction if I declare it in the `updater` package. Should I put it in the `internal` repo or is what I have fine?

Duplicating the divertRequest is fine for now :)

> 3\. I'm assuming we don't need a fully backwards- and forwards-compatible migration path for the changes in each of the 3 affected repos and can just commit (I think) the `gokrazy` and `updater` changes first, then update `go.mod` in `tools` to point at the updated dependencies.

No, we actually do need compatibility in both directions. Some people might update their gok cli, and still expect it to work with gokrazy instances that haven’t been updated ot the latest github.com/gokrazy/gokrazy yet. On the other end of the spectrum, some folks might have module updates automated (as described on https://gokrazy.org/userguide/automation/) and will get a newer github.com/gokrazy/gokrazy, but still have an older gok cli.

In this case, it’s quite easy, though: 

1. check for the `Content-Type` header to be `application/json`. If that’s not the case, fill a divertRequest from r.FormValue() calls.
2. On the gokrazy/updater side, fall back to request parameters when you get a BadRequest HTTP response status code.

markdrayton:
> No, we actually do need compatibility in both directions.

Gah, in hindsight that is very obvious :-(

Updated PRs coming shortly.

markdrayton:
Okay, here are the updated PRs. I have tested this in the various combinations (new `gok`/gokrazy, old `gok`/new gokrazy, new `gok`/old gokrazy and it seems to do the job. The only thing I wondered a little about was including a warning to the user if they try to pass arguments to a program under `gok run` on a version of gokrazy that doesn't support passing them. It might be a little confusing to include arguments that appear to be ignored.

stapelberg:
> The only thing I wondered a little about was including a warning to the user if they try to pass arguments to a program under `gok run` on a version of gokrazy that doesn't support passing them

Probably not too many users will run into it, but yeah, it would be nice. I think the best way to accomplish said error is by changing https://github.com/gokrazy/updater/pull/1 to return an error from the fallback path if len(flags) > 0

update.go
```diff
@@ -394,23 +394,39 @@ func uploadTemp(w http.ResponseWriter, r *http.Request) {
...
 	if r.Method != http.MethodPost {
 		http.Error(w, "expected a POST request", http.StatusBadRequest)
 		return
 	}
 
-	diversion := r.FormValue("diversion")
-	if diversion == "" {
+	var req divertRequest
+	if r.Header.Get("Content-Type") == "application/json" {
+		err := json.NewDecoder(r.Body).Decode(&req)
```
stapelberg, line 411: This is a common mistake: json.NewDecoder “reads and decodes JSON values from an input stream.”, but we don’t have an input stream — we only have a single value. It’s better to just call json.Unmarshal() on the result of io.ReadAll(resp.Body) in this case.

stapelberg (commented):
Looks good overall!

update.go
markdrayton: Fixed!

stapelberg (approved):

PR #177, opened 2023-03-05
implement Poweroff (and refactor around killSupervisedServices)

### LINKED ISSUE #176: Proper way to poweroff
## Platform

* [x] Raspberry Pi Model B Rev 2 (https://github.com/gokrazy-community/kernel-rpi-os-32)

## Observed behavior

There is no way to shut down the raspberry pi.
I only found a way to reboot, by POSTing to `/reboot`.

## Expected behavior

I would like to shutdown my raspberry pi properly (to switch the SD card for instance) by issuing a dedicated request (POST `poweroff` or `halt` or `shutdown`?)

Even better would be to have a button on the https interface for both operations (rebooting, shutting down).

## Context

I had to replace my SD (with was defective) and I want to preserve the new SD card as much as possible :)

I would be ready to draft a PR (I guess [LINUX_REBOOT_CMD_POWER_OFF](https://man7.org/linux/man-pages/man2/reboot.2.html) is the right syscall)

oliverpool:
Fixes #176 

I took the liberty of refacoring `killSupervisedServices` on the way:
- it sends SIGTERM to all running processes
- waits up to 15 seconds (`=signalDelay`)
- sends SIGKILL to the remaining processes
- waits up to 15 seconds (`=signalDelay`)
- returns

So that we get a higher chance of all processes having been terminated before unmount `/perm`.

(I initially replaced `syscall.Kill` with `process.Signal` and proper error checking, however rebooting always took 30s. After I while, I figured that `process.Signal` does not send the signal to the group! I reverted my changes and updated the comment :)

I have no idea how to hook this to the power button: I will let you take it from here

### DIFF
--- a/supervise.go
+++ b/supervise.go
@@ -286,13 +286,9 @@ func (s *service) Signal(signal syscall.Signal) error {
 	s.processMu.RLock()
 	defer s.processMu.RUnlock()
 	if s.process != nil {
-		// Use syscall.Kill instead of s.process.Signal since we want
-		// to the send the signal to all process of the group (-pid)
-		err := syscall.Kill(-s.process.Pid, signal)
-		if errno, ok := err.(syscall.Errno); ok {
-			if errno == syscall.ESRCH {
-				return nil // no such process, nothing to signal
-			}
+		err := s.process.Signal(signal)
+		if err != nil && !errors.Is(err, os.ErrProcessDone) { // ignore "process not found"
+			return err
 		}
 		return err
 	}
--- a/supervise.go
+++ b/supervise.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"log/syslog"
 	"net/http"
@@ -287,15 +286,11 @@ func (s *service) Signal(signal syscall.Signal) error {
 	s.processMu.RLock()
 	defer s.processMu.RUnlock()
 	if s.process != nil {
-		// Use syscall.Kill instead of s.process.Signal as the latter returns
-		// “os: process already finished” when the process exited.
-		err := syscall.Kill(-s.process.Pid, signal)
-		if errno, ok := err.(syscall.Errno); ok {
-			if errno == syscall.ESRCH {
-				return nil // no such process, nothing to signal
-			}
+		err := s.process.Signal(signal)
+		if err != nil && !errors.Is(err, os.ErrProcessDone) {
+			return err
 		}
-		return err
+		return nil
 	}
 	return nil // no process, nothing to signal
 }
@@ -329,7 +324,7 @@ func (s *service) MarshalJSON() ([]byte, error) {
 }
 
 func rssOfPid(pid int) int64 {
-	statm, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/statm", pid))
+	statm, err := os.ReadFile(fmt.Sprintf("/proc/%d/statm", pid))
 	if err != nil {
 		return 0
 	}
@@ -354,7 +349,7 @@ func (s *service) RSS() int64 {
 var syslogRaddr string
 
 func initRemoteSyslog() {
-	b, err := ioutil.ReadFile("/perm/remote_syslog/target")
+	b, err := os.ReadFile("/perm/remote_syslog/target")
 	if err != nil {
 		if !os.IsNotExist(err) {
 			log.Print(err)
--- a/supervise.go
+++ b/supervise.go
@@ -534,13 +534,15 @@ var services struct {
 	S []*service
 }
 
-// killSupervisedServices is called before rebooting when upgrading, allowing
-// processes to terminate in an orderly fashion.
-func killSupervisedServices() {
+// signalSupervisedServices sends a given signal to all non-stopped processes.
+// It returns the corresponding processState to allow waiting for a given state.
+func signalSupervisedServices(signal syscall.Signal) []*processState {
 	services.Lock()
 	defer services.Unlock()
+
+	states := make([]*processState, 0, len(services.S))
 	for _, s := range services.S {
-		if s.Stopped() {
+		if s.Stopped() && s.state.Get() == Stopped {
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


supervise.go
```diff
@@ -539,13 +538,15 @@ var services struct {
...
+// signalSupervisedServices sends a given signal to all non-stopped processes.
+// It returns the corresponding processState to allow waiting for a given state.
+func signalSupervisedServices(signal syscall.Signal) []*processState {
 	services.Lock()
 	defer services.Unlock()
+
+	states := make([]*processState, 0, len(services.S))
 	for _, s := range services.S {
-		if s.Stopped() {
+		if s.Stopped() && s.state.Get() == Stopped {
```
stapelberg, line 549: Can you add a comment explaining why checking Stopped() is not enough in this case?

supervise.go
```diff
@@ -287,8 +286,8 @@ func (s *service) Signal(signal syscall.Signal) error {
 	s.processMu.RLock()
 	defer s.processMu.RUnlock()
 	if s.process != nil {
-		// Use syscall.Kill instead of s.process.Signal as the latter returns
-		// “os: process already finished” when the process exited.
+		// Use syscall.Kill instead of s.process.Signal since we want
+		// to the send the signal to all process of the group (-pid)
 		err := syscall.Kill(-s.process.Pid, signal)
```
stapelberg, line 291: Thinking about it, I think we should use s.process.Signal() after all. 

Killing the process group might be too aggressive: for example when running a program that runs qemu VMs: qemu doesn’t power off the VM, but exits directly when receiving SIGTERM.

supervise.go
```diff
@@ -554,12 +555,49 @@ func killSupervisedServices() {
...
+		states = append(states, s.state)
 	}
+	return states
+}
 
-	for _, s := range services.S {
-		s.waitTillStopped()
+// killSupervisedServices is called before rebooting when upgrading, allowing
+// processes to terminate in an orderly fashion.
+func killSupervisedServices(signalDelay time.Duration) {
```
stapelberg, line 566: I know I said in the issue that there should be a time after which shutdown should proceed. I had forgotten about this existing SIGTERM code, which blocks until the process exits. 

I think that’s actually the better model (at least by default), as it exposes the hanging shutdown to the user (ideally during development when it’s obvious where the problem comes from) who is then incentivized to fix it :)

Can you adjust the code to keep waiting indefinitely please?

stapelberg (commented):
Thanks, this looks good overall! Just a couple of small adjustments please :)

supervise.go
stapelberg: Oh, I just now realize that the 15 seconds timeout is already in the existing code. Nevertheless, I think it might be good to make this timeout configurable. On the machine that runs my VM, I would want to configure it way higher than just 15 seconds (as the VM might take a while to shut down, too).

supervise.go
oliverpool: Added

```
		// s.Stopped() only checks the "stopped" flag of the service (if it shouldn't restart).
		// We check the actual state as well to be sure to re-send a signal if we are already
		// in the "Stopping" state.
```

supervise.go
stapelberg: > But then we loose the signal in case of privilege dropping (like ntp is doing for instance).

Hmm. ntp should probably either handle the signal by forwarding it to the privdrop process, or set [SysProcAttr.Pdeathsig](https://pkg.go.dev/syscall#SysProcAttr) so that the child receives SIGTERM itself when the parent exits.

I think other init systems don’t kill the entire pgid. systemd for example only sends the signal to the main process.

> I think this is more of an issue with qemu (which should properly exit on SIGTERM).

It should indeed, but the lack of signal handling in qemu is an open bug for many years: https://gitlab.com/qemu-project/qemu/-/issues/148

> Or can you change the PGID of qemu after starting it? This way your original process gets the signal and can properly communicate with qemu to shut down

I probably can, but do we really want to require all processes to start children in a separate pgid to avoid sending SIGTERM to children? I think only sending the signal to the main process is probably the less surprising choice.

stapelberg:
> I have no idea how to hook this to the power button: I will let you take it from here

Okay! I just wrote the code which reads power button events, so I’m ready to hook this up once this PR is merged :)

supervise.go
oliverpool: For now looking at the logs, it is as-if the sigterm and sigkill were simply ignored by grafana (or ntp before the update).

oliverpool:
So regarding the GroupPID, the decision to be made is:

## keep terminating the process group (implementation before this PR)

The advantage is that existing code (like `ntp` or `grafana-armv6`) will keep working as-is. No dangling processes (since they will all be cleaned on group sigkill at the latest

## Or switch to terminating only the parent process (current implementation or this PR)

The qemu layer does not have to worry about creating a separated process group, since it will not get the sigterm signal.

---

Since both options always have a workaround for the other case, I think it sums up to what is more usual.

I think I would personally prefer the current state (signal to group), since I find it more intuitive and I do not have a usecase where a subprocess should not get the sigterm while the parents gets it (I have examples the other way around).

@stapelberg what do you think? Maybe we could ping other gokrazy users to ask about their usecases?

stapelberg:
Thanks for the updates!

> I think I would personally prefer the current state (signal to group), since I find it more intuitive and I do not have a usecase where a subprocess should not get the sigterm while the parents gets it (I have examples the other way around).

It’s a good point that with the kill-pgid approach, it is the least amount of code for `exec.Command()` users, which is a nice property :)

I compared what systemd-shutdown is doing, and it also sends SIGTERM to all running (non-root) processes, so that is the expected behavior. Implementing that on gokrazy by killing the pgid sounds like a reasonable way to both send SIGTERM to every supervised process and children, but also to not be overzealuous (= send SIGTERM to **all** processes, no matter if supervised or not), which would prevent the escape hatch of creating a new process group when using exec.Command().

So, yes, please change it back to killing the pgid, and I’ll change my qemu runner.

Thanks

supervise.go
```diff
@@ -554,12 +550,49 @@ func killSupervisedServices() {
...
+	termStates := signalSupervisedServices(syscall.SIGTERM)
```
stapelberg, line 562: Can you add a log message before sending SIGTERM, too, please?

update.go
```diff
@@ -331,42 +331,73 @@ func initUpdate() error {
...
+		killSupervisedServicesAndUmountPerm(signalDelay)
+		fmt.Fprintf(w, "All services killed in %s\n", time.Since(start))
+		rc.Flush()
 
-			log.Println("Unmounting /perm")
-			if err := syscall.Unmount("/perm", unix.MNT_FORCE); err != nil {
-				log.Printf("unmounting /perm failed: %v", err)
-			}
+		log.Println("Rebooting")
+		if err := reboot(); err != nil {
```
stapelberg, line 351: Can you keep the reboot() call in a separate goroutine here? The current code breaks the gokrazy updater, because it doesn’t complete the HTTP request before rebooting.

stapelberg (changes requested):

supervise.go
oliverpool: Added `sending sigterm to all services`

oliverpool:
I reverted the changes and now perform the reboot/poweroff in a dedicated goroutine.

stapelberg (approved):

stapelberg:
Thank you!

PR #178, opened 2023-03-05
doc: how to make changes to the init process

oliverpool:
It took me quite some time to figure out that the init process was not influenced by modifying the `~/gokrazy/instancename/builddir/github.com/gokrazy/gokrazy/go.mod` but by `~/gokrazy/instancename/builddir/init/go.mod`.

This PR adds a short paragraph at the end of the readme to hopefully prevent others making the same mistake :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Turns out this is feasible. gokrazy is the result.
 
 # GitHub Repository structure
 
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for sending this! 

Another contributor ran into a similar issue recently.

Maybe it would make sense to change the code so that the builddir is no longer init, but github.com/gokrazy/gokrazy as you expected. I’ll try and see if that’s possible.

PR #179, opened 2023-03-06
dashboard: link to the raw log

oliverpool:
Currently the number of lines in the dashboard is hard-coded at 101.
This is sometimes not enough (I want to debug victoria metrics: a panic yields the stacktrace for tens of gorountines, which don't fit in 101 lines).

This PR introduces an `input` at the right of the log name, to adjust the number of log lines. This is set to 101 by default and gets considered everytime a new log line is produced. I tried to add some JS optimization along the way.

What do you think?

### DIFF
--- a/internal/assets/status.tmpl
+++ b/internal/assets/status.tmpl
@@ -43,10 +43,10 @@
     </p>
     {{ end }}
 
-    <h3>stdout</h3>
+    <h3>stdout <small class="pull-right">Log lines: <input type="number" value="101" size="4" id="stdout_len"></small></h3>
     <pre id="stdout"></pre>
 
-    <h3>stderr</h3>
+    <h3>stderr <small class="pull-right">Log lines: <input type="number" value="101" size="4" id="stderr_len"></small></h3>
     <pre id="stderr"></pre>
 
     <h3>module info</h3>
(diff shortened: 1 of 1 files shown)


oliverpool:
After thinking about this, maybe instead of an input I could add a link to the log: https://gokrazy/log?path=/user/service-name&stream=stderr

This way you get the raw log (with new entries as soon as they are added). That way you don't have to guess how many log lines are needed.

What do you think?

damdo:
+1 for the link to the raw logs

oliverpool:
I replaced the form with a simple `raw log` link:

stapelberg:
We already have the `gok logs` command. Maybe we should just print a corresponding command line (`gok -i <instance> logs --service <service>`) instead of a link?

oliverpool:
I prefer having a URL with the log, than having to use `gok` for that:
- my `gok` instantiation is already quite painful because I also have to specify a parent_dir
- having the full log just a click away is very nice (and seeing it in the browser gives smooth scrolling, quick search and filesave for free - yes, those would be achievable in the CLI, but you have to know in advance that you need them...)

stapelberg:
> * having the full log just a click away is very nice (and seeing it in the browser gives smooth scrolling, quick search and filesave for free - yes, those would be achievable in the CLI, but you have to know in advance that you need them...)

Ok, that’s fair.

Can you split up the PR to do only the URL, and split the truncation change into a separate PR please? I can quickly merge the URL change, but need to work through the truncation change when I have more time.

> * my `gok` instantiation is already quite painful because I also have to specify a parent_dir

Do you know about the `GOKRAZY_PARENT_DIR` environment variable? That way, you don’t need to specify the flag on every invocation :)

oliverpool:
Superseded by #180 & #181

PR #181, opened 2023-03-12
dashboard: refactor log line truncation logic

oliverpool:
Following #179 

I think this is the fastest logic without introducing a new state (which would keep in memory the number of lines, to prevent having to count from 0 each time).

### DIFF
--- a/internal/assets/status.tmpl
+++ b/internal/assets/status.tmpl
@@ -57,7 +57,8 @@
 
 {{ template "footer.tmpl" . }}
 
-<script>const maxLines = 101;
+<script>
+    const maxLines = 101;
     function newLogrotate(elt) {
       var lines = 0;
       return function (e) {
--- a/internal/assets/status.tmpl
+++ b/internal/assets/status.tmpl
@@ -58,8 +58,8 @@
 {{ template "footer.tmpl" . }}
 
 <script>
-    const maxLines = 101;
     function newLogrotate(elt) {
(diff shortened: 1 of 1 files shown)


internal/assets/status.tmpl
```diff
@@ -58,16 +58,37 @@
 {{ template "footer.tmpl" . }}
 
 <script>
+    // truncateAt finds the position at which the prefix should be removed
+    // to have the right number of lines (for use with slice).
+    function truncateAt(s, maxLines) {
+      let found = 0;
+      let i = s.length + 1;
+      while (true) {
+        i = s.lastIndexOf("\n", i - 1);
```
stapelberg, line 67: When displaying 100 lines, this is a 100 step approach.

I would prefer if we would:

1. count the number of lines we’re adding
2. remove this number of lines from innerText — the number of lines we’re adding will be much lower than the maximum number of lines (in innerText)

That should make it even faster. Can you give that a try please?

oliverpool: I have added a `lines` counter ("introducing a new state" from my first message).

And on https://jsben.ch/ it showed an even bigger improvement.
I unfortunately lost the result, but from what I remember:
- my initial version was 15-20% faster than the original
- this version is 100% faster than the original (2x as fast)

stapelberg (approved):

stapelberg:
Thank you!

PR #182, opened 2023-03-12
dashboard: show latest stdout and stderr lines

oliverpool:
The current logic showing the "last log line" on the dashboard is flawed if sorting the log lines does not result in a chronological order.

Example with an empty stdout and the following stderr:
```
Hello world
A beautiful day
```

The current code sorts the lines and takes the last element. In my example, it returns `Hello world`, which is unexpected.

https://github.com/gokrazy/gokrazy/blob/ac1c5aaf638d8de41aca7c234c1c09ab61b040df/status.go#L146-L153

I have tried to fix the logic, but choosing between stderr and stdout will be lottery (as long as the timestamp is not properly recorded).

Hence I resorted to showing both, when present:

I find the result quite good, since most services are only using stderr (and when both are present, no that much space is wasted).

What do you think?

### DIFF
--- a/internal/assets/overview.tmpl
+++ b/internal/assets/overview.tmpl
@@ -23,13 +23,18 @@
 {{ end }}
 </td>
 <td class="lastlog">
-{{ last $svc.Stdout.Lines $svc.Stderr.Lines }}
+{{ if $svc.Stdout.Lines }}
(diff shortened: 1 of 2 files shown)


stapelberg (approved):

stapelberg:
Good idea! Thanks :)

PR #187, opened 2023-03-23
dhcp: wait indefinitely for interface to appear

### LINKED ISSUE #144: DHCP client started for non-existant eth0 on Pi Zero series
## Platform

I’m using:

* [ ] gokrazy/rpi3b
* [x] gokrazy/rpi3b+
* [ ] gokrazy/rpi4b
* [ ] gokrazy/apu2c4
* [ ] gokrazy/x86-64
* [x] gokrazy-community/rpi-zero 

## Observed behavior

The DHCP service is always set to autostart unless a `dontstart.txt` is explicitly provided for it.
The Pi Zero series do not have a wired ethernet interface, therefor the process will constantly crash and restart.

## Expected behavior

The DHCP service could see there is, in fact, no eth0, and exit cleanly. (Maybe wait a little bit in case the user actually plugged in a wired USB ethernet adapter)

piger:
On systems without an ethernet interface (like the Raspberry Pi Zero) the `dhcp` program goes into a crash loop because it can't find eth0.

This PR adds an initial loop to wait for the given interface name to appear before continuing with the rest of the program.

**NOTE**: Unfortunately I can't fully test this code because I don't have any ethernet dongle; on a plain Raspberry Pi Zero 2 W it works fine: the `dhcp` process is not crash looping anymore.

Fixes #144

### DIFF
--- a/cmd/dhcp/dhcp.go
+++ b/cmd/dhcp/dhcp.go
@@ -277,6 +277,18 @@ func deprioritizeRoutesWhenDown(nl *netlink.Handle, ifname string, extraRoutePri
 	}
 }
 
+func waitForInterface(ifname string) {
+	t := time.NewTicker(1 * time.Second)
+	defer t.Stop()
+
+	log.Printf("waiting indefinitely for %s to appear", ifname)
+	for range t.C {
+		if _, err := net.InterfaceByName(ifname); err == nil {
+			return
+		}
+	}
+}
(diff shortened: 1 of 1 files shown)


piger:
I've bought an ethernet dongle and ran a second test: once the device is plugged in `dhcp` "unblocks" itself and run as expected.

PR #192, opened 2023-04-04
make Model detect QEMU, always return non-empty string

bradfitz:
And cache its result, as it's called regularly by at least fbstatus, which seems like is using too much CPU. (though admittedly the fbstatus CPU problems are mostly elsewhere)

And rename mustReadFile0 because a "must" prefix implies it panics.

### DIFF
--- a/status.go
+++ b/status.go
@@ -14,6 +14,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync/atomic"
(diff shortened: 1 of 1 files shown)


bradfitz:
cc @stapelberg

stapelberg (approved):

stapelberg:
Thank you!

PR #198, opened 2023-04-17
fix waitforclock condition when creating a new service

dtchanpura:
WaitForClock condition doesn't make sense to have on a Stopped service.

This change in behavior happened due to [this commit](https://github.com/gokrazy/gokrazy/commit/7e14fb215c2487a0507124e742f847dac22e1df2) which fixes the panic.


stapelberg (approved):

stapelberg:
Thanks!

PR #206, opened 2023-07-14
Add waiw to "Written for gokrazy"

BrunoTeixeira1996:
I use this project daily to store movies that I watch with my gf but since I have a gokrazy instance up and running I decided to make it possible to run in gokrazy too.
This works easily by just using a postgresql service running somewhere outside the gokrazy env.
Decided to make this as an example to the `Written for gokrazy` as this works with and without gokrazy.

### DIFF
--- a/website/content/userguide/rsync-backups.markdown
+++ b/website/content/userguide/rsync-backups.markdown
@@ -77,11 +77,7 @@ dont_namespace = true
 EOT
 ```
 
-(Note that `scan2drive.lan` is the name of my gokrazy instance. I added the IP of the instance and defined a name in `/etc/hosts`, i.e `192.168.10.2 scan2drive.lan`)
-
 Finally, create the authorized keys file that determines who can access the rsync daemon:
-
-(i.e `ssh-keygen -t ed25519 -C "michael@midna.com"`, save in `~/.ssh/id_ed25519_scan2drivebackup` and cat the `.pub` key to `gokr-rsyncd.authorized_keys`)
 
 ```bash
 cat > ~/gokrazy/hello/gokr-rsyncd.authorized_keys <<'EOT'
--- a/website/content/userguide/rsync-backups.markdown
+++ b/website/content/userguide/rsync-backups.markdown
@@ -77,7 +77,11 @@ dont_namespace = true
 EOT
 ```
 
+(Note that `scan2drive.lan` is the name of my gokrazy instance. I added the IP of the instance and defined a name in `/etc/hosts`, i.e `192.168.10.2 scan2drive.lan`)
+
 Finally, create the authorized keys file that determines who can access the rsync daemon:
+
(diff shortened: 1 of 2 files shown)


BrunoTeixeira1996:
[#4f1](https://github.com/gokrazy/gokrazy/pull/206/commits/4f1cbd0dcd3ad2079060186e562f5fda65dffcd9) was about making rsync documentation better.

website/content/userguide/rsync-backups.markdown
```diff
@@ -77,8 +77,12 @@ dont_namespace = true
 EOT
 ```
 
+(Note that `scan2drive.lan` is the name of my gokrazy instance. I added the IP of the instance and defined a name in `/etc/hosts`, i.e `192.168.10.2 scan2drive.lan`)
```
stapelberg, line 80: No, that’s not needed in my network. The hostnames that devices send to my router when they configure an IP address using DHCP is available for name resolution using DNS. This is a pretty common setup for home networks, but not all routers support it (some require extra settings).

I think we should just adjust this example to use `backedup` as the hostname so that it matches the `config.json` above, and no extra explanation — the concept of a hostname should be clear to gokrazy users.

website/content/userguide/rsync-backups.markdown
```diff
@@ -77,8 +77,12 @@ dont_namespace = true
...
+
 Finally, create the authorized keys file that determines who can access the rsync daemon:
 
+(i.e `ssh-keygen -t ed25519 -C "michael@midna.com"`, save in `~/.ssh/id_ed25519_scan2drivebackup` and cat the `.pub` key to `gokr-rsyncd.authorized_keys`)
```
stapelberg, line 84: I don’t think this line is good to add — it repeats what the command below it already says.

You don’t need to generate an ed25519 key specifically (it was the default on my machine), and the comment was likewise auto-generated. 

Creating an authorized keys file should be familiar to SSH users. If you want, we can add a link to a tutorial (or documentation) regarding SSH public keys and authorized_keys files, but I don’t want to spell it out in terms of “run ssh-keygen with these options”

stapelberg (changes requested):

website/content/userguide/rsync-backups.markdown
BrunoTeixeira1996: Ok sure I understand!

BrunoTeixeira1996: @stapelberg if you want just approve my other commit about `waiw`  and lets not make changes on the docs

stapelberg: I don’t know how to approve only a single commit. Can you change the PR to only contain one commit please?

BrunoTeixeira1996: @stapelberg there you go!

stapelberg (approved):

PR #207, opened 2023-07-17
Change how status is displayed

Syphdias:
Currently the status is appended to the service status link and no label
is displayed if the service is running. The status stopped is displayed
with a warning color (yellow). To me this feels uneven and a bit chaotic
which makes it harder to get a good overview and find a service.

This does three things to improve the overview:
- Add a separate column for status label. This takes a bit of width away
  from the last log line.
- Color the stopped label less aggressive (gray/default instead of
  yellow/warning).
- Add a green label for running service.

Potential improvements for the future could be to show if a service is
enabled/supervised or adding actions like start/stop/restart/enable.

### DIFF
--- a/internal/assets/overview.tmpl
+++ b/internal/assets/overview.tmpl
@@ -8,18 +8,23 @@
 <table class="table">
 <tbody><tr>
 <th width="20%">path</th>
-<th width="80%">last log line</th>
+<th width="8%">status</th>
+<th width="72%">last log line</th>
 </tr>
 
 {{ range $idx, $svc := .Services }}
 <tr>
 <td>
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for your PR! Could you include a screenshot of the updated interface please?

Syphdias:
Will do, as soon as I get home today.

Syphdias:
Before:
image

After:
image


PS: Both versions have the same behavior that a restarting service that gets stopped is still in restarting. Not sure if this is intentional or not.

stapelberg (approved):

stapelberg:
Thank you!

PR #234, opened 2023-12-10
docs: adds GPIO input example

### LINKED ISSUE #233: GPIO / periph.io Edge detection / sysfs  not working
Hi! 👋 

Thanks for this awesome project! ❤️ 

## Platform

I’m using:

* [ ] gokrazy/rpi3b
* [ ] gokrazy/rpi3b+
* [ ] gokrazy/rpi4b
* [x] gokrazy/rpiz2w
* [ ] gokrazy/apu2c4
* [ ] gokrazy/x86-64

I've followed the [GPIO guide](https://gokrazy.org/development/gpio/) and was able to run the example (Low, High, Low etc).

I then proceeded to configure my GPIO Pin for Input, which works fine if  I use no rising edge:

```go
if err := p.In(gpio.PullNoChange, gpio.NoEdge); err != nil {
	log.Fatal(err)
}
fmt.Printf("%s is %s\n", p, p.Read())
```

However, since I want my program to react on inputs/events, I was hoping to be able to use edge detection as per the [periph.io button docs](https://periph.io/device/button/), unfortunately there seem to be some issue setting up/accessing the `/sys/class/gpio` path. I've used breakglass and there is no such path.

I can see that sysfs is enabled in the gokrazy kernel, so I'm a bit at a loss here what to do.

Found this comment on this depedency:
https://github.com/periph/host/blob/f673315ab3a9ebe15db798b450d193bc7a9a7f1c/bcm283x/gpio.go#L373-L380

Code that is failing:

```go
if err := p.In(gpio.PullNoChange, gpio.RisingEdge); err != nil {
	log.Fatal(err)
}
fmt.Printf("%s is %s\n", p, p.Read())

for b.Pin.WaitForEdge(10 * time.Millisecond) {
	fmt.Printf("%s went %s\n", b.Pin, gpio.High)
}
```

## Observed behavior

```shell
2023/12/09 20:22:26 Loading periph.io drivers
2023/12/09 20:22:26 bcm283x-gpio (GPIO4): pin 4 is not exported by sysfs
2023/12/09 20:22:26 gokrazy: exit status 1
```

## Expected behavior

Able to use GPIO edge detection as per `periph.io` docs.

eripa:
Feel free to update the Go code as I'm not sure if this is the best approach, but I've confirmed that it's working for my use-case and is performing reasonably well.

This PR also updates to latest `v3` `periph.io` module structure for the GPIO output example. 

Resolves #233

### DIFF
--- a/website/content/development/gpio.md
+++ b/website/content/development/gpio.md
@@ -10,7 +10,7 @@ In this guide, we are using [periph.io](https://periph.io/), a library for
 peripheral I/O in Go, to set one of the Raspberry Pi’s General Purpose I/O
 (GPIO) pins to a logical high (3.3V) or low (0V) signal.
 
-periph.io supports the Raspberry Pi 3 and Raspberry Pi 4, starting with version
+periph.io supports the Raspberry Pi 3, Raspberry Pi Zero 2W and Raspberry Pi 4, starting with version
 `v3.6.4`.
 
 ## Connect GPIO pins based on pinout
@@ -38,9 +38,9 @@ import (
 	"log"
 	"time"
 
-	"periph.io/x/periph/conn/gpio"
-	"periph.io/x/periph/host"
-	"periph.io/x/periph/host/rpi"
+	"periph.io/x/conn/v3/gpio"
+	host "periph.io/x/host/v3"
+	"periph.io/x/host/v3/rpi"
 )
 
 func doGPIO() error {
@@ -59,7 +59,6 @@ func doGPIO() error {
 		}
 		<-t.C
 	}
-	return nil
 }
 
 func main() {
--- a/website/content/development/gpio.md
+++ b/website/content/development/gpio.md
@@ -84,3 +84,95 @@ At this point, we should be able to see the high/low signal on the multimeter,
 alternating between 3.3V (high) and 0V (low) every 5 seconds:
 
 <a href="/img/2020-06-15-gpio.jpg"><img src="/img/2020-06-15-gpio.thumb.jpg" srcset="/img/2020-06-15-gpio.thumb.2x.jpg 2x,/img/2020-06-15-gpio.thumb.3x.jpg 3x" width="700" style="border: 1px solid grey; margin-bottom: 2em; margin-top: 1em"></a>
+
+## Setting up an input pin
+
+To setup a Pin as an input and to act on change to High level, such as a button,
+we must poll the Pin periodically because [periph.io’s](http://periph.io/) edge
+detection, which examples like the
+[periph.io/device/button example](https://periph.io/device/button/) rely on,
+[is not available on gokrazy](https://github.com/gokrazy/gokrazy/issues/233).
+
+The below example configures a single Pin for Input and uses Go channels to
+signal level changes that can be acted on. Keep in mind that shortening the
+polling interval will increase CPU usage.
+
+```go
+package main
+
+import (
+	"log"
+	"time"
+
+	"periph.io/x/conn/v3/gpio"
+	"periph.io/x/conn/v3/gpio/gpioreg"
+	host "periph.io/x/host/v3"
+)
+
+type pinLevelMessage struct {
+	State gpio.Level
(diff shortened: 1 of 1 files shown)


website/content/development/gpio.md
```diff
@@ -85,3 +84,101 @@ At this point, we should be able to see the high/low signal on the multimeter,
 alternating between 3.3V (high) and 0V (low) every 5 seconds:
 
 <a href="/img/2020-06-15-gpio.jpg"><img src="/img/2020-06-15-gpio.thumb.jpg" srcset="/img/2020-06-15-gpio.thumb.2x.jpg 2x,/img/2020-06-15-gpio.thumb.3x.jpg 3x" width="700" style="border: 1px solid grey; margin-bottom: 2em; margin-top: 1em"></a>
+
+## Setting up an input pin
+
+To setup a Pin as an input and to act on change to High level, such as a button,
+we must poll the Pin periodically. This is needed because the
```
stapelberg, line 91: How about:

…we must poll the Pin periodically because periph.io’s edge detection, which examples like the [periph.io/device/button example](https://periph.io/device/button/) rely on, is [not available on gokrazy](https://github.com/gokrazy/gokrazy/issues/233).

website/content/development/gpio.md
```diff
@@ -85,3 +84,101 @@ At this point, we should be able to see the high/low signal on the multimeter,
...
+import (
+	"log"
+	"time"
+
+	"periph.io/x/conn/v3/gpio"
+	"periph.io/x/conn/v3/gpio/gpioreg"
+	host "periph.io/x/host/v3"
+)
+
+type PinLevelMessage struct {
```
stapelberg, line 112: This type doesn’t need to be exported. Exports in the main package don’t really do anything, either, but I think it’s good style to export as little as possible. So, change to lowercase: `type pinLevelMessage struct{`

website/content/development/gpio.md
```diff
@@ -85,3 +84,101 @@ At this point, we should be able to see the high/low signal on the multimeter,
...
+
+	// Setup Input signalling
+	go func() {
+		lastLevel := p.Read()
+		// How often to poll levels, 100-150ms is fairly responsive unless
+		// button presses are very fast.
+		// Shortening the polling interval <100ms significantly increases
+		// CPU load.
+		t := time.NewTicker(125 * time.Millisecond)
+		for {
```
stapelberg, line 140: This can just be: `for range time.Tick(125*time.Millisecond) {`

website/content/development/gpio.md
```diff
@@ -85,3 +84,101 @@ At this point, we should be able to see the high/low signal on the multimeter,
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Main loop, act on level changes
+	for {
+		// How often to check for channel changes, generally it is fine to have
+		// a slightly longer interval here because the signal change is already
+		// caught at this point and will be waiting in the channel.
+		t := time.NewTicker(200 * time.Millisecond)
```
stapelberg, line 168: Why this additional level of delay? You could just read from levelChan blockingly, no?

stapelberg (changes requested):
Thanks for updating the docs!

website/content/development/gpio.md
eripa: Sounds better, thanks 😊

website/content/development/gpio.md
eripa: Of course, just an oversight. Thanks

website/content/development/gpio.md
eripa: TIL, correct. Thanks!

website/content/development/gpio.md
eripa: Not sure why I put this in here.. possible before I could get the blocking channel to work 🤔. I've removed it and tested it w/o, works fine!

eripa:
Thanks @stapelberg, pushed an updated commit.

stapelberg (approved):

PR #241, opened 2024-01-07
dhcp: Allow -static_network_config to read from a file

bdd:
It might be more desirable to deploy the static network configuration as its own JSON file instead of needing to deal with string quoting it to fit into `config.json`.

- Add a simple heuristic for checking if the flag value is a JSON string.
  - If the value is JSON, preserve the original behavior
  - If not, treat it value as a file path and load the JSON from there.

- Set lease source string to reflect the source of static configuration.

### DIFF
--- a/cmd/dhcp/dhcp.go
+++ b/cmd/dhcp/dhcp.go
@@ -368,8 +368,24 @@ func main() {
 	}
 
 	if *staticConfig != "" {
+		var sc []byte
+		var lsrc string
+
+		// Simple heuristic to check if we got JSON strings from the command line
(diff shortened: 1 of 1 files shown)


stapelberg (approved):

stapelberg:
Thanks!

damdo:
Thanks! Probably worth adding an entry to the docs for it :)

stapelberg:
> Thanks! Probably worth adding an entry to the docs for it :)

Yeah. Done with commit https://github.com/gokrazy/gokrazy/commit/ab7a3e3f04b05f8a67edc664f366f880b31b93ba

PR #244, opened 2024-01-17
ntp: Allow alternative NTP servers as arguments

bdd:
When the gokrazy deployment doesn't have internet access and the network lacks NTP traffic re-routing to the local time servers, instances may either have downright wrong clock (e.g. Raspberry Pi with no RTC) or experience divergence due to drift.

This add the ability to override the default NTP Pool Vendor Zone for gokrazy with servers supplied as command line arguments.

e.g.
```
    "PackageConfig": {
        "github.com/gokrazy/gokrazy/cmd/ntp": {
            "CommandLineFlags": [
                "time1.local" "time2.local" "10.1.2.123"
            ]
        }
    }
```

### DIFF
--- a/cmd/ntp/ntp.go
+++ b/cmd/ntp/ntp.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	"flag"
 	"log"
 	"math/rand"
 	"os"
@@ -39,6 +40,13 @@ func set(rtc *os.File) error {
 
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
+
+	flag.Parse()
+
+	if len(flag.Args()) > 0 {
+		servers = flag.Args()
+		log.Printf("using command line supplied server list: %v", servers)
+	}
 
 	var rtc *os.File
 	var err error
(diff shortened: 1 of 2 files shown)


cmd/ntp/ntp.go
```diff
@@ -40,6 +40,10 @@ func set(rtc *os.File) error {
 func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
 
+	if len(os.Args) > 1 {
+		servers = os.Args[1:]
```
stapelberg, line 44: Just to make sure that we can add flags in the future, could you change this function to call `flag.Parse()`, then use `flag.Args()`?

stapelberg (changes requested):

cmd/ntp/ntp.go
bdd: Ah. Much better.

stapelberg: I don’t see a change yet? Did you forget to force-push the branch maybe?

bdd: I did not push yet as I decided to also add a log line to show NTP servers in use.
It's coming.

bdd: deciding the 6 words logging string took ~15 minutes of going through alternatives and not liking any. please change the wording as you see it.

stapelberg (approved):

PR #251, opened 2024-02-21
Add instructions to the main page to create the permanent storage filesystem in macOS.

drio:
Screenshot 2024-02-21 at 09 27 18

I tested this in my system and works perfectly.

### DIFF
--- a/website/content/quickstart/_index.markdown
+++ b/website/content/quickstart/_index.markdown
@@ -27,8 +27,8 @@ go install github.com/gokrazy/tools/cmd/gok@main
 
 ## Step 1: Insert an SD card
 
-Run `watch -d1 ls -l '/dev/disk/by-id/*'` (macOS: `diskutil list`) and insert an SD card. Copy the
-highlighted device name:
+Run `watch -d1 ls -l '/dev/disk/by-id/*'` (see the permanent [section](https://gokrazy.org/userguide/permanent-data/section) for macOS instructions)
+and insert an SD card. Copy the highlighted device name:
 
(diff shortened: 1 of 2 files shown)


damdo:
I think that rather than adding a macOS "guide" it'd be more future proof to remove these steps altogether and instead link https://gokrazy.org/userguide/permanent-data/ (which also refers to gokrazy/mkfs) where we could keep things up to date more easily and for different build platforms.

drio:
It makes sense. 
I'll modify the PR to add a section in https://gokrazy.org/userguide/permanent-data/. 
Should we still have a link in the main page storage section to https://gokrazy.org/userguide/permanent-data?

stapelberg (changes requested):
Thanks for your PR!

drio (commented):
Thank you for the 👀. Let me know if you want me to change anything else.

damdo:
Personally now that we are at it, I'd also change the sentences at https://github.com/gokrazy/gokrazy/blob/7fca4d5ddd109192500abdf8dcf700ce119594f9/internal/assets/overview.tmpl#L79-L82
in favour of a link something along the lines of:
```
No permanent storage mounted. To create a filesystem for permanent storage follow the steps at https://gokrazy.org/userguide/permanent-data/
```
But let's see what @stapelberg thinks

stapelberg (approved):

stapelberg:
> But let's see what @stapelberg thinks

Linking the docs sounds good, but I’m not entirely sure whether we should remove the mkfs command. Can you send that as a separate PR please?

PR #256, opened 2024-03-13
website: update tailscale package page

joneskoo:
Mention how to use gokrazy as a Tailscale subnet router and how to sign in with Tailscale SSH.

### DIFF
--- a/website/content/packages/tailscale.md
+++ b/website/content/packages/tailscale.md
@@ -157,7 +157,7 @@ than your gokrazy appliance, you can set up Tailscale as a subnet router.
 }
 {{< /highlight >}}
 
-Starting Tailscale version v1.63.x ([FIXME](https://github.com/tailscale/tailscale/issues/11405))
+Starting [Tailscale version v1.64](https://github.com/tailscale/tailscale/issues/11405),
 IP forwarding is automatically enabled on Gokrazy.
 
 ## Optional: Accessing gokrazy using Tailscale SSH
--- a/website/content/packages/tailscale.md
+++ b/website/content/packages/tailscale.md
@@ -133,6 +133,61 @@ instance to not require login every 3 months.
 [Tailscale console]: https://login.tailscale.com/ "Tailscale management console login.tailscale.com"
 {{% /notice %}}
 
+## Optional: Using gokrazy as a Tailscale subnet router
+
+To allow hosts in your Tailnet to access devices in your network other
+than your gokrazy appliance, you can set up Tailscale as a subnet router.
+
+{{< highlight json "hl_lines=12" >}}
+{
+    "Hostname": "ts",
+    "Packages": [
+        "github.com/gokrazy/serial-busybox",
+        "tailscale.com/cmd/tailscaled",
+        "tailscale.com/cmd/tailscale"
+    ],
+    "PackageConfig": {
+        "tailscale.com/cmd/tailscale": {
+            "CommandLineFlags": [
+                "up",
+                "--advertise-routes=192.168.0.0/24"
+            ]
+        }
+    }
+}
+{{< /highlight >}}
+
+Starting Tailscale version v1.63.x ([FIXME](https://github.com/tailscale/tailscale/issues/11405))
+IP forwarding is automatically enabled on Gokrazy.
+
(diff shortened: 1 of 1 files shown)


joneskoo:
I'm wondering whether it'd make sense to also replace the go listener with `tailscale serve` as that's now a nice way to do similar thing for many use cases.

Also, the new authorization features could be nice for more complex use cases, but on the other hand there's no particular reason to document that explicitly in gokrazy.

It might be interesting to find out and document how to make gokrazy only accessible through Tailscale. Maybe even replace basic authentication with Tailscale? :)

website/content/packages/tailscale.md
```diff
@@ -133,6 +133,61 @@ instance to not require login every 3 months.
...
+            "CommandLineFlags": [
+                "up",
+                "--advertise-routes=192.168.0.0/24"
+            ]
+        }
+    }
+}
+{{< /highlight >}}
+
+Starting Tailscale version v1.63.x ([FIXME](https://github.com/tailscale/tailscale/issues/11405))
```
stapelberg, line 160: It’s been a while without progress on this PR. How do you want to proceed? Do you want to split this PR into two so that we can merge the SSH part right now? Or document a small program to set the required sysctl?

website/content/packages/tailscale.md
```diff
@@ -133,6 +133,61 @@ instance to not require login every 3 months.
...
+            "CommandLineFlags": [
+                "up",
+                "--ssh=true"
+            ]
+        }
+    }
+}
+{{< /highlight >}}
+
+Note that Tailscale SSH needs to be allowed by your Tailnet ACL.
```
stapelberg, line 187: Could you add a link to the relevant tailscale docs here?

stapelberg:
> I'm wondering whether it'd make sense to also replace the go listener with `tailscale serve` as that's now a nice way to do similar thing for many use cases.

Yeah, we should probably document `tailscale serve` more prominently and the Go listener less prominently (if at all). I’m still learning how to best run Tailscale myself… :) 

> It might be interesting to find out and document how to make gokrazy only accessible through Tailscale. Maybe even replace basic authentication with Tailscale? :)

An interesting thought. When you say “replace basic authentication”, can you outline how removing it entirely is better than using fixed credentials? I.e., you could configure user gokrazy, password gokrazy. This frees you from managing per-device passwords. The only downside I can see is that you still have a password prompt in the web browser. 

So maybe removing HTTP basic auth for tailscale requests from the web interface is the minimal change we’d need to make to get all the convenience benefits? Or did I miss something here?

joneskoo:
Moved the replace basic authentication to https://github.com/gokrazy/gokrazy/issues/257

website/content/packages/tailscale.md
joneskoo: I'm hopeful Tailscale will merge a version of the PR I made as Brad Fitzpatrick suggested it'd be likely reasonable for Tailscale to do that on gokrazy. Too bad it's taking longer than I thought to get triaged.

Could we instruct how to pick the version from the PR? I think a separate program for sysctl would work as well but as I'm already using the version of tailscale from my PR, I wouldn't need it. If you think such example program serves some more general example, then that would make sense.

stapelberg: > Could we instruct how to pick the version from the PR? I think a separate program for sysctl would work as well but as I'm already using the version of tailscale from my PR, I wouldn't need it. If you think such example program serves some more general example, then that would make sense.

I’d rather not. If the PR won’t get merged in its current form, users are left in the rain. Maybe it’s best to just hold off on this PR for now until the other one moves.

joneskoo:
I expect this would be included in v1.63 as it was merged to main in tailscale but I suppose we should wait for the release?

joneskoo:
@stapelberg Tailscale skipped v1.63 and v1.64.0 was tagged and includes the forwarding feature.

drio:
They should have proudly mentioned the gokrazy related adition in the changelog. ☝🏻

Screenshot 2024-04-11 at 17 22 05

stapelberg (approved):

stapelberg:
Great, thanks everyone!

PR #266, opened 2024-06-06
Perform poweroff instead of halt for 32-bit arch

anupcshan:
LINUX_REBOOT_CMD_HALT is higher than int32 MAX and doesn't compile as an int on 32-bit systems. Ideally, unix.Reboot would take a uint.

### DIFF
--- a/halt.go
+++ b/halt.go
@@ -1,0 +1,17 @@
+//go:build !arm
+// +build !arm
+
+package gokrazy
+
+import (
+	"log"
+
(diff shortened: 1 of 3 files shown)


stapelberg:
Thanks for sending this PR! I think commit b48fd2647893ae0d5322d6842d0607584e21cd55 is the better fix, though. Let me know if that works on your device.

anupcshan:
Can confirm USR1 and USR2 work as expected

PR #275, opened 2024-08-12
Support specifying remote_syslog target via Linux cmdline

bradfitz:


bradfitz:
cc @stapelberg

supervise.go
```diff
@@ -355,15 +355,33 @@ func (s *service) RSS() int64 {
...
+		return strings.TrimSpace(string(b))
+	}
+	if !os.IsNotExist(err) {
+		log.Print(err)
+	}
+
+	// As another option, support reading the target value from
+	// the Linux command line. This is intended for use in
+	// integration tests where Linux is being booted from a qemu
+	// microvm, booting the kernel directly.
```
stapelberg, line 370: I’m not opposed to this approach, but I wonder if falling back from /perm/remote_syslog/target to /etc/gokrazy/remote_syslog/target would work just as well for your use-case? 

In general, I think I’m converging on reading config from /etc (i.e. include it in the gokrazy instance config.json, either as a literal or an extra file), with /perm being the override for it. As an example, if you use `gok new`, the config.json it creates will contain the authorized_files for breakglass in /etc, as well as a unique machine id. 

You do control the file system in your setup, yes?

bradfitz: I could tolerate that. It's a bit less ideal because we'd need a specific Gokrazy instance for integration testing vs the one end users use, but maybe that's not terrible anyway.

Should /etc or /perm be higher priority?

You want me to update this PR just for `remote_syslog/target` or for more config items?

bradfitz: Actually, I just found a case where putting it in /etc is annoying: when we run our connectivity integration tests in an IPv6-only configuration, we can't hard-code a fake IPv4 address in the image. So then to change the network topology, we need to rebuild the image. I suppose I could use DNS for it, though.

stapelberg: > Should /etc or /perm be higher priority?

/perm should be higher priority (so that one can change settings at runtime without rebuilding a new instance)

> You want me to update this PR just for `remote_syslog/target` or for more config items?

Just for the syslog target for now. Do you have others in mind that could use the same treatment?

bradfitz:
@stapelberg, sorry, I lost this PR and just found it back. Updated. PTAL?

/cc @cmol (who noticed we lost this support during a recent rebase)

stapelberg (approved):

PR #290, opened 2025-01-19
Document GOKRAZY_PARENT_DIR

### LINKED ISSUE #222: Document 'GOKRAZY_PARENT_DIR' environment variable
I'm probably not the only person who doesn't want to have to use `--parent_dir` to specify a location each time we run `gok` because we don't want ( yet another ) directory created by an application in our home directory. For example, I have a directory for code projects: `~/Code`, so I wanted to change the default parent directory from `~/gokrazy` to `~/Code/GoKrazy`.

I spent a long time searching to see if `gok` will optionally read from a config file in `~/.config/gok` or something, before eventually finding the `GOKRAZY_PARENT_DIR` environment variable. Huzzah! 

It would be nice if this was easier to find in the documentation -- or if there **is** a config file we could create in `~/.config/gok` if that could be documented instead.

vilhalmer:
As a non-conformist, I needed to move my instances elsewhere and had to discover this variable via the code. :)

I explicitly added just the relevant parts of what Hugo generated, regenerating the entire site touches a lot of files due to changes in generation in the latest releases. If you do want to do the full rebuild I'm happy to change this PR.

The current release is also unhappy with the site in general:
```
ERROR deprecated: .Site.IsMultiLingual was deprecated in Hugo v0.124.0 and will be removed in Hugo 0.142.0. Use hugo.IsMultilingual instead.
```

I'm not sure how/if you'd like to handle this; I can open another issue if you want to track upgrading, or update the README to specify a version. Either way I figured it's out of scope for this PR.

Fixes #222.

### DIFF
--- a/website/content/quickstart/_index.markdown
+++ b/website/content/quickstart/_index.markdown
@@ -44,6 +44,9 @@ gok new
 # alternatively, to use a different instance name: gok -i myname new
 ```
 
+The `--parent_dir` can also be changed permanently by setting the
+`GOKRAZY_PARENT_DIR` environment variable.
+
 If you’re curious, you can run `gok edit` to open the generated `config.json` in
 your editor.
 
--- a/docs/quickstart/index.html
+++ b/docs/quickstart/index.html
@@ -845,7 +845,9 @@ and insert an SD card. Copy the highlighted device name:</p>
 <div class="highlight"><pre tabindex="0" style="color:#f8f8f2;background-color:#272822;-moz-tab-size:4;-o-tab-size:4;tab-size:4;"><code class="language-bash" data-lang="bash"><span style="display:flex;"><span>gok new
 </span></span><span style="display:flex;"><span><span style="color:#75715e"># creates an instance named “hello” in --parent_dir default directory ~/gokrazy</span>
 </span></span><span style="display:flex;"><span><span style="color:#75715e"># alternatively, to use a different instance name: gok -i myname new</span>
-</span></span></code></pre></div><p>If you’re curious, you can run <code>gok edit</code> to open the generated <code>config.json</code> in
+</span></span></code></pre></div><p>The <code>--parent_dir</code> can also be changed permanently by setting the
+<code>GOKRAZY_PARENT_DIR</code> environment variable.</p>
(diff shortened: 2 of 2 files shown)


website/content/quickstart/_index.markdown
```diff
@@ -47,6 +47,9 @@ gok new
 If you’re curious, you can run `gok edit` to open the generated `config.json` in
 your editor.
 
+If you prefer to store all of your instances elsewhere, the parent directory can
```
stapelberg, line 50: Line 43 already talks about the parent dir and shows how to change it (per invocation). Can you add a line after line 45 instead? Maybe something like:

```
# To permanently use a different --parent_dir, set the GOKRAZY_PARENT_DIR environment variable.
```

docs/index.json
```diff
@@ -4,7 +4,7 @@
...
...ted web interface on private IP addresses\n(each instance gets its own password in the Update → HTTPPassword key of its config.json, see gok edit) supervise all installed programs (only the hello world program in this example) To interact with your device, you can:\nOpen the gokrazy web interface in your browser (easiest option) Attach a serial console (see below) Use the breakglass package to start an interactive SSH session. gok new sets it up by default. Congratulations! 🎉 You now have a working gokrazy installation!\nNext, see the Configuration guide for details on how to use other programs.\nOptional: Using the serial console See also: Configuration → Instance Config Reference → SerialConsole\nIf you prefer the serial console, you can optionally connect a serial adapter to the Raspberry Pi and remove the \u0026quot;SerialConsole\u0026quot;: \u0026quot;disabled\u0026quot; line from your gokrazy instance’s config.json to make the primary Linux console end up on the serial console.\nWhen pressing Enter on the serial console, gokrazy will interactively start either:\n/tmp/serial-busybox/ash (provided by the serial-busybox package) /perm/sh, if present (supplied by you, the user) "
```
stapelberg, line 7: Can you revert your change to this index file please? I’ll update it as part of the website regeneration step.

stapelberg (changes requested):
Thanks for your PR!

vilhalmer:
Done and done!

PR #296, opened 2025-02-09
Always listen on the GokrazyHTTPUnixSocket

tho:
The change is made to allow other tools like [breakglass](https://github.com/gokrazy/breakglass) or [mkfs](https://github.com/gokrazy/mkfs) to interact with the Gokrazy API when TLS is enabled. The tools currently make HTTP requests to get system information or trigger reboots. When TLS is enabled, these requests fail because Gokrazy redirects all HTTP requests to HTTPS. While it is possible to configure the respective clients to properly validate (self-signed) certificates, doing so is cumbersome. Therefore, always listen on the Unix socket to avoid the complexity of certificate validation. Username and password must still be provided and will be validated.

I've considered removing the username and password validation for socket requests. However, I've decided to reverse the change and keep everything as it is for now, in part to keep changes to a minimum. I can see arguments for and against removing the checks, and I don't feel strongly one way or the other.

The `GokrazyHTTPUnixSocket` constant is exported so that tools can reference it instead of hardcoding the path.

Ref https://github.com/gokrazy/breakglass/issues/21

### DIFF
--- a/listeners.go
+++ b/listeners.go
@@ -43,8 +43,8 @@ var (
 	listeners   = make(map[string][]*krazyServer) // by unix socket path and private IP address
 )
 
-// GokrazyHTTPUnixSocket is the unix socket path that the HTTP server listens on.
-const GokrazyHTTPUnixSocket = "/run/gokrazy-http.sock"
+// HTTPUnixSocket is the unix socket path that the HTTP server listens on.
+const HTTPUnixSocket = "/run/gokrazy-http.sock"
 
 // tlsConfig: tlsConfig. nil, if the listeners should not use https (e.g. for redirects)
 func updateListeners(port, redirectPort string, tlsEnabled bool, tlsConfig *tls.Config) error {
@@ -58,7 +58,7 @@ func updateListeners(port, redirectPort string, tlsEnabled bool, tlsConfig *tls.
 	// If NoPassword is used, the HTTP server doesn't run over HTTP
 	// and instead only listens on a Unix socket. Packages running
 	// in the appliance can proxy to the Unix socket as desired.
-	hosts := []string{GokrazyHTTPUnixSocket}
+	hosts := []string{HTTPUnixSocket}
 	if httpPassword != "" {
 		addrs, err := PrivateInterfaceAddrs()
 		if err != nil {
@@ -120,7 +120,7 @@ func updateListeners(port, redirectPort string, tlsEnabled bool, tlsConfig *tls.
 		}
 		// add a new listener
 		var srv *http.Server
-		if tlsEnabled && tlsConfig == nil && host != GokrazyHTTPUnixSocket {
+		if tlsEnabled && tlsConfig == nil && host != HTTPUnixSocket {
 			// "Redirect" server
 			srv = &http.Server{
 				Handler:   http.HandlerFunc(httpsRedirect(redirectPort)),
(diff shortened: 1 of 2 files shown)


listeners.go
```diff
@@ -39,28 +40,31 @@ type krazyServer struct {
...
 	listenersMu sync.Mutex
-	listeners   = make(map[string][]*krazyServer) // by private IP address or unix socket path
+	listeners   = make(map[string][]*krazyServer) // by unix socket path and private IP address
 )
 
-// gokrazyHTTPUnixSocket is the unix socket path that
-// the HTTP server listens on when httpPassword is empty.
-const gokrazyHTTPUnixSocket = "/run/gokrazy-http.sock"
+// GokrazyHTTPUnixSocket is the unix socket path that the HTTP server listens on.
+const GokrazyHTTPUnixSocket = "/run/gokrazy-http.sock"
```
stapelberg, line 47: I also wonder if we should make this `/run/gokrazy/http.sock` (so that we can add other files later without littering the top-level /run), or maybe even `/run/gokrazy/api.sock` (more descriptive than `http`).

@bradfitz Are you okay with changing this path? Do you have stuff running that will break if we change it?

listeners.go
```diff
@@ -39,28 +40,31 @@ type krazyServer struct {
...
 var (
 	listenersMu sync.Mutex
-	listeners   = make(map[string][]*krazyServer) // by private IP address or unix socket path
+	listeners   = make(map[string][]*krazyServer) // by unix socket path and private IP address
 )
 
-// gokrazyHTTPUnixSocket is the unix socket path that
-// the HTTP server listens on when httpPassword is empty.
-const gokrazyHTTPUnixSocket = "/run/gokrazy-http.sock"
+// GokrazyHTTPUnixSocket is the unix socket path that the HTTP server listens on.
```
stapelberg, line 46: Can you rename this constant to HTTPUnixSocket please? The package is already named gokrazy, which the current name repeats.

listeners.go
```diff
@@ -39,28 +40,31 @@ type krazyServer struct {
...
+const GokrazyHTTPUnixSocket = "/run/gokrazy-http.sock"
 
 // tlsConfig: tlsConfig. nil, if the listeners should not use https (e.g. for redirects)
 func updateListeners(port, redirectPort string, tlsEnabled bool, tlsConfig *tls.Config) error {
-	var hosts []string // private IP addresses or unix socket path
-	var err error
-
+	// Always listen on a Unix socket.
+	//
+	// Local packages running as root can connect through the Unix socket
```
stapelberg, line 53: Change “packages” to “processes” please. You are correct in that gokrazy deals with Go packages, but at the level we are talking about here (runtime), the distinction is executable files that get loaded into operating system processes :)

listeners.go
```diff
@@ -39,28 +40,31 @@ type krazyServer struct {
...
+	// regardless of TLS settings. With TLS enabled: Avoids the complexity
```
stapelberg, line 54: I’m not sure the TLS settings deserve this much mention in the comment. How about:

```
// Local processes running as root can connect through the Unix socket,
// which removes the complexity of connecting via HTTP+TCP
// (which might need a custom port, or TLS credentials).
```

stapelberg (changes requested):
Thanks, looks good overall!

stapelberg (approved):

PR #319, opened 2025-06-02
Add optional async argument to /reboot & /poweroff

damdo:
At the moment HTTP calls to /reboot and /poweroff are blocking and need to wait for `killSupervisedServicesAndUmountPerm` to finish before returning a positive response to the caller.

This can get in the way when the caller is a supervised service running on the same gorkrazy instance.
As the caller waits for the HTTP request to get a response back, and the HTTP handler waits for all the services to be terminated. This can cause a deadlock until the supervised service is SIGKILLed after the configured deadline.

This PR introduces an async agrument that can be passed in the HTTP call as an argument (e.g. `/reboot?async=true`), which allows the HTTP handler to ack the request and return 200 OK, whilst performing the sigterm of the supervised processes (including the caller process) asynchronously.
 
Context: https://github.com/gokrazy/selfupdate/pull/4#discussion_r2116201013

Example:

```
# with default behaviour, sync reboot
$ curl -X POST "http://gokrazy:bla@127.0.0.1:8080/reboot"
All services killed in 3.72021348s
Rebooting...

# with arg specified but not true, sync reboot
$ curl -X POST "http://gokrazy:bla@127.0.0.1:8080/reboot?async=false"
All services killed in 3.710093377s
Rebooting...

# with arg specified and true, immediate async reboot
$ curl -X POST "http://gokrazy:bla@127.0.0.1:8080/reboot?async=true"
Rebooting...
```

### DIFF
--- a/update.go
+++ b/update.go
@@ -331,75 +331,63 @@ func initUpdate() error {
 		}
 	})
 	http.HandleFunc("/reboot", func(w http.ResponseWriter, r *http.Request) {
-		if r.Method != http.MethodPost {
-			http.Error(w, "expected a POST request", http.StatusBadRequest)
-			return
-		}
-
-		signalDelay := defaultSignalDelay
-		if s := r.FormValue("wait_per_signal"); s != "" {
-			var err error
-			signalDelay, err = time.ParseDuration(s)
-			if err != nil {
-				http.Error(w, err.Error(), http.StatusBadRequest)
-				return
-			}
-		}
-
-		rc := http.NewResponseController(w)
-		start := time.Now()
-		killSupervisedServicesAndUmountPerm(signalDelay)
-		fmt.Fprintf(w, "All services killed in %s\n", time.Since(start))
-		rc.Flush()
-
-		log.Println("Rebooting")
-		w.Write([]byte("Rebooting...\n"))
-		rc.Flush()
-
-		go func() {
-			time.Sleep(time.Second) // give the http response some time to be sent
-			if err := reboot(r.FormValue("kexec") != "off"); err != nil {
-				log.Println("could not reboot:", err)
-			}
-		}()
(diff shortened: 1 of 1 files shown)


update.go
```diff
@@ -345,10 +345,18 @@ func initUpdate() error {
...
+		if r.FormValue("async") == "true" {
+			log.Println("Asynchronous reboot requested")
+			go func() {
+				killSupervisedServicesAndUmountPerm(signalDelay)
+			}()
+		} else {
+			log.Println("Synchronous reboot requested")
+			start := time.Now()
+			killSupervisedServicesAndUmountPerm(signalDelay)
+			fmt.Fprintf(w, "All services killed in %s\n", time.Since(start))
```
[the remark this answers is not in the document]
damdo: Done, in the async case I had to convert it to a log.Printf instead of a fmt.Fprintf to avoid writing to the `http.ResponseWriter` after the handler has already returned

update.go
```diff
@@ -378,10 +388,18 @@ func initUpdate() error {
...
 		rc := http.NewResponseController(w)
-		start := time.Now()
-		killSupervisedServicesAndUmountPerm(signalDelay)
-		fmt.Fprintf(w, "All services killed in %s\n", time.Since(start))
-		rc.Flush()
+
+		if r.FormValue("async") == "true" {
+			log.Println("Asynchronous power off requested")
+			go func() {
+				killSupervisedServicesAndUmountPerm(signalDelay)
```
stapelberg, line 395: This branch is still missing the “All services killed” message. In fact, maybe we should take this opportunity to refactor these identical lines into a shared function, so that the two places cannot go out of sync?

stapelberg (changes requested):

update.go
damdo: Oh it looks like I need to rebase. Will do it tomorrow

Issue #320, opened 2025-06-07
Feature Request: Automatically set the $USER env variable


jboverfelt:
Hi there - I've had great success running gokrazy, thanks for making it!

On one of my appliances (an rpi 4b), I run a DLNA/UPNP server (https://github.com/anacrolix/dms) with the following gokrazy config:

```json
        "github.com/anacrolix/dms": {
            "CommandLineFlags": [
                "--friendlyName=media-server",
                "--path=/mnt/share",
                "--noProbe",
                "--noTranscode",
                "--ifname=eth0"
            ],
            "Environment": [
                "USER=root"
            ],
            "WaitForClock": true
        }
```

Since this package uses Go's os/user and we're on Linux without cgo, os/user expects to have $USER and $HOME defined. Gokrazy already defines $HOME, but I had to set $USER manually (to root since everything seems to run as root unless you explicitly drop privs). 

I can certainly keep setting $USER in the Environment config but I was wondering if this is common enough that you'd consider setting $USER by default. I'd offer a patch but I don't know how you'd like to approach it (set to root every time? something else?).

PR #322, opened 2025-06-07
Automatically set USER to root if not already set

jboverfelt:
I included a check to make sure we don't overwrite something that's already set. We don't seem to do that for $HOME, but I think users are much more likely to want to override $USER than $HOME.

Ref #320

### DIFF
--- a/supervise.go
+++ b/supervise.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


jboverfelt:
Hi there! Have you had a chance to review this PR?

stapelberg (approved):

stapelberg:
Sorry for the late reply, the last few weeks were very busy with travel and work. I just now got a chance to try out your PR, and it seems to work fine :) Thanks!

jboverfelt:
Totally understand! Thanks Michael!

PR #331, opened 2025-09-06
web interface: add power management buttons

oliverpool:
This PR adds a basic menu at the top right of the web interface to easily shutdown gokrazy (useful to prevent SD-Card corruption when unplugging the board).
image
image


damdo:
I think it is nice to have that.
Tough I wonder if we can think of other places where we could put this instead.
What happens when you click the Power Management collapsed menu?
Does it open and increase the height of the header pushing the whole UI down vertically?

stapelberg:
> Nitpick; That is what I tought. I think I'd be nice to find a place and a way where this does not happen :)

Yeah, it’s not super elegant, but the entire header has grown so much that it would probably be good to clean it up altogether at some point. For now, I’ll merge this change as-is.

stapelberg (approved):

PR #359, opened 2026-01-11
cmd/ntp: preserve system time in /perm.

fischman:
Why? Reduce the wall time during which the system believes it's 1970.
Makes syslog timestamps more sensible.

### DIFF
--- a/cmd/ntp/ntp.go
+++ b/cmd/ntp/ntp.go
@@ -3,20 +3,34 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"io"
 	"log"
 	"math/rand"
 	"os"
+	"os/signal"
 	"syscall"
 	"time"
 
 	"github.com/beevik/ntp"
 )
 
+const cookiePath = "/perm/.ntp-time-at-last-shutdown"
+
 var servers = []string{
 	"0.gokrazy.pool.ntp.org",
 	"1.gokrazy.pool.ntp.org",
 	"2.gokrazy.pool.ntp.org",
 	"3.gokrazy.pool.ntp.org",
+}
+
+func setTimeOfDay(t time.Time, source string) error {
+	tv := syscall.NsecToTimeval(t.UnixNano())
+	if err := syscall.Settimeofday(&tv); err != nil {
+		return fmt.Errorf("syscall.Settimeofday(%v): %v", tv, err)
+	}
+	log.Printf("clock set to %v (from %s)", t, source)
+	return nil
 }
 
 func set(rtc *os.File) error {
@@ -26,16 +40,46 @@ func set(rtc *os.File) error {
 		return err
 	}
 
-	tv := syscall.NsecToTimeval(r.Time.UnixNano())
-	if err := syscall.Settimeofday(&tv); err != nil {
-		return err
+	if err := setTimeOfDay(r.Time, server); err != nil {
+		return fmt.Errorf("setTimeOfDay: %v", err)
 	}
-	log.Printf("clock set to %v (from %s)", r.Time, server)
 
 	if rtc == nil {
 		return nil
 	}
 	return setRTC(rtc, r.Time.UTC())
+}
+
+func loadTime(cookie *os.File) error {
+	buf, err := io.ReadAll(cookie)
+	if err != nil {
+		return fmt.Errorf("io.ReadAll(%v): %v", cookiePath, err)
+	}
+	var t time.Time
+	if err := t.UnmarshalText(buf); err != nil {
+		return fmt.Errorf("t.UnmarshalText(%v): %v", string(buf), err)
+	}
+	if err := setTimeOfDay(t, cookiePath); err != nil {
+		return fmt.Errorf("setTimeOfDay: %v", err)
+	}
+	return nil
+}
+
+func saveTime(cookie *os.File) error {
+	buf, err := time.Now().MarshalText()
+	if err != nil {
+		return fmt.Errorf("time.Now().MarshalText: %v", err)
+	}
+	if err := cookie.Truncate(0); err != nil {
+		return fmt.Errorf("cookie.Truncate: %v", err)
+	}
+	if _, err := cookie.Seek(0, 0); err != nil {
+		log.Printf("cookie.Seek(0, 0): %v", err)
+	}
+	if _, err := cookie.Write(buf); err != nil {
+		return fmt.Errorf("cookie.Write(%v): %v", buf, err)
+	}
+	return nil
 }
 
 func main() {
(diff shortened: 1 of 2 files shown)


fischman:
@stapelberg this is slightly speculative. I like the resulting cleanliness of my /perm/syslog/<instance> directory & logs. I slightly worry that people are {ab,}using WaitForClock as a proxy for WaitForNetwork and will be unhappy if the clock is set before the network is available. LMK what you think.

cmd/ntp/ntp.go
```diff
@@ -3,41 +3,85 @@ package main
...
 }
 
+func loadTime(cookie *os.File) error {
+	buf, err := io.ReadAll(cookie)
+	if err != nil {
+		return fmt.Errorf("io.ReadAll(%v): %v", cookiePath, err)
+	}
+	var t time.Time
+	if err := t.UnmarshalText(buf); err != nil {
+		return fmt.Errorf("t.UnmarshalText(%v): %v", string(buf), err)
```
stapelberg: I think this should be “time.UnmarshalText” instead of “t” — because for “t”, you need to look at the source to understand what type it is.

cmd/ntp/ntp.go
```diff
@@ -3,41 +3,85 @@ package main
...
+	if err := cookie.Truncate(0); err != nil {
+		return fmt.Errorf("cookie.Truncate: %v", err)
+	}
+	if _, err := cookie.Seek(0, 0); err != nil {
+		log.Printf("cookie.Seek(0, 0): %v", err)
+	}
+	if _, err := cookie.Write(buf); err != nil {
+		return fmt.Errorf("cookie.Write(%v): %v", buf, err)
+	}
+	return nil
```
stapelberg, line 85: Let’s also close the file here, just to be sure the contents are written out by the OS.

cmd/ntp/ntp.go
```diff
@@ -48,27 +92,61 @@ func main() {
...
 	}
 
-	mustDropPrivileges(rtc)
+	if cookie != nil {
+		// Save current time at shutdown.
+		go func() {
+			// Wait for SIGTERM to save time to /perm.
+			ch := make(chan os.Signal, 1)
+			signal.Notify(ch, syscall.SIGTERM)
+			defer func() {
```
stapelberg: Instead of re-sending the signal, let’s just directly `os.Exit(128+syscall.SIGTERM)` after the `saveTime` call.

cmd/ntp/ntp.go
```diff
@@ -48,27 +92,61 @@ func main() {
...
+				log.Printf("persisting time to /perm failed: %v", err)
+			}
+		}()
+		// Load time saved at previous shutdown, if any.
+		if err := loadTime(cookie); err != nil {
+			log.Printf("loadTime: %v", err)
+		}
+	}
 
+	jitter := time.Duration(rand.Int63n(250)) * time.Millisecond
```
stapelberg: Why the change to jitter? Seems unrelated to preserving the time?

cmd/ntp/ntp.go
```diff
@@ -3,41 +3,85 @@ package main
...
 	"math/rand"
 	"os"
+	"os/signal"
 	"syscall"
 	"time"
 
 	"github.com/beevik/ntp"
 )
 
+const cookiePath = "/perm/.ntp-time-at-last-shutdown"
```
stapelberg: I think we should use a more descriptive name than cookie for this file. How about “timefile”?

Also, let’s not hide this file. The ntp process is started by gokrazy in `/perm/home/ntp`, so you can just use “timefile” as a path and it will end up in the right place and be visible should users need to delete or inspect it.

stapelberg (changes requested):
Seems reasonable on a high level. Left a couple of detail comments.

stapelberg:
> slightly worry that people are {ab,}using WaitForClock as a proxy for WaitForNetwork and will be unhappy if the clock is set before the network is available. LMK what you think.

I don’t worry about that. If people use the mechanism wrong, they can fix their usage. 

In general, we do want to make changes backwards-compatible. But, we don’t need to go out of our way in terms of supporting **broken** setups, I think.

cmd/ntp/ntp.go
fischman: Renamed the variable, dropped leading period from value basename, and dropped directory path.
I didn't realize that gokrazy would "backfill" a /perm if partition 4 wasn't a writeable filesystem.

cmd/ntp/ntp.go
fischman: lol that's what I tried initially, only to find that package time doesn't contain a free UnmarshalText function :)
(lmk if you see one I'm missing)

cmd/ntp/ntp.go
fischman: Done.

(I actually went on a side-quest wondering about gokrazy's reboot semantics vis-a-vis (the lack of) sync(2) calls, before realizing only /perm is writable, and it gets umount'ed during shutdown/reboot, forcing a sync(2))

cmd/ntp/ntp.go
fischman: Done.

cmd/ntp/ntp.go
fischman: A perhaps non-obvious change in this PR is that previously, failure to set time from NTP would sleep 0-250ms and then exit the process, to get respawned by the supervisor, and repeat until success.
Contrarily, in this PR the repeated attempts to use NTP successfully for the first time will happen in a single process, and I couldn't see a reason for varying the jitter between those calls.
Further, varying jitter between hourly check-ins to NTP also seemed strange.

I'm fine with reverting to a per-iteration jitter value if you think that's worthwhile.

stapelberg: I’d like to leave the jitter behavior as-is, i.e. vary it between hourly checkins.

The reason for jitter between repeated calls is to prevent the thundering herd problem for the NTP pool operators. The reason to vary jitter between calls is that the randomness might be rather similar on different Raspberry Pis that boot up without a clock source or network yet. So my concern is that if we picked one specific jitter value, the “herd” of gokrazy devices might end up with the same jitter, thereby defeating the concept. Varying between calls circumvents this problem.

cmd/ntp/ntp.go
stapelberg: Oh, sorry, to be clear, I meant just in the error message, not in the code :)

cmd/ntp/ntp.go
```diff
@@ -3,41 +3,88 @@ package main
...
 	}
-	log.Printf("clock set to %v (from %s)", r.Time, server)
 
 	if rtc == nil {
 		return nil
 	}
 	return setRTC(rtc, r.Time.UTC())
 }
 
+func loadTime(cookie *os.File) error {
```
stapelberg: Here and elsewhere, please also rename `cookie` to `timefile`. We can name the constant (used less frequently) `timefilepath`.

stapelberg (changes requested):

cmd/ntp/ntp.go
fischman: Done.

cmd/ntp/ntp.go
fischman: Done.

cmd/ntp/ntp.go
fischman: Done, and added explanatory comment.

FTR I am skeptical of the concern here because I can't fathom enough Pi's sharing a power circuit booting all at once to overwhelm an NTP server. But it's your project and I don't feel strongly about this :)

stapelberg (approved):

PR #369, opened 2026-03-16
Handle routes for off subnet gateways

JustinAzoff:
If the gateway is not within the host's subnet (e.g. with a /32 address), add a host route for the gateway on the interface first, otherwise adding the default route will fail.  This is a popular setup on cloud providers.  Often DHCP Option 121 will be provided to indicate this route is necessary, but this simpler check is enough for the common scenario.

### DIFF
--- a/cmd/dhcp/dhcp.go
+++ b/cmd/dhcp/dhcp.go
@@ -186,6 +186,24 @@ func applyLease(nl *netlink.Handle, ifname string, source string, lease dhcp4.Le
 	}
 
 	if r := lease.Router; len(r) > 0 {
+		// If the gateway is not within the host's subnet (e.g. with a
+		// /32 address), add a host route for the gateway on the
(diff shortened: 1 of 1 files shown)


JustinAzoff:
With this in place a /32 default route gets added properly:
```
2026/03/16 12:28:48 dhcp.go:160: DHCPACK: IP 172.30.0.7/32, router 172.30.0.1, DNS [1.1.1.1 9.9.9.9]
```
```
/ # ip route
default via 172.30.0.1 dev eth0  metric 1 
172.30.0.1 dev eth0 scope link  metric 1 
/ # 
```

JustinAzoff:
If you think it would be better, I can also take a stab at adding full support for dhcp option 121 here and to rtr7/dhcp4. The main difference is that will change the options requested in the initial discover to include DHCPOptClasslessStaticRoute and parsing of the option in the response.

PR #376, opened 2026-06-04
docs(containers): update for podman 5+ with netavark

damdo:
Update docs for podman 5+ with netavark/nftables

### DIFF
--- a/website/content/packages/docker-containers.markdown
+++ b/website/content/packages/docker-containers.markdown
@@ -18,11 +18,9 @@ re-package into https://github.com/gokrazy/podman.
 Include the following packages in your gokrazy installation:
 
 ```bash
-gok add github.com/gokrazy/iptables
 gok add github.com/gokrazy/nsenter
 gok add github.com/gokrazy/podman
-gok add github.com/greenpau/cni-plugins/cmd/cni-nftables-portmap
-gok add github.com/greenpau/cni-plugins/cmd/cni-nftables-firewall
+gok add github.com/spod/nftables
 ```
 
 Then, deploy as usual:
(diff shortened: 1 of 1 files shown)


website/content/packages/docker-containers.markdown
```diff
@@ -18,11 +18,9 @@ re-package into https://github.com/gokrazy/podman.
 Include the following packages in your gokrazy installation:
 
 ```bash
-gok add github.com/gokrazy/iptables
 gok add github.com/gokrazy/nsenter
 gok add github.com/gokrazy/podman
-gok add github.com/greenpau/cni-plugins/cmd/cni-nftables-portmap
-gok add github.com/greenpau/cni-plugins/cmd/cni-nftables-firewall
+gok add github.com/spod/nftables
```
stapelberg, line 23: Why doesn’t Podman configure nftables natively, e.g. with the https://pkg.go.dev/github.com/google/nftables package?

stapelberg (approved):

website/content/packages/docker-containers.markdown
damdo: Yeah good point, unfortunately it needs it:
```
Error: netavark: nftables error: unable to execute "nft": No such file or directory (os error 2)
```

Maybe the podman folks could be open to change this?

stapelberg: Worth asking in an upstream issue, I think. Tailscale is entirely self-contained and does not shell out to `nft`, for example :)

damdo: Ah ofc netavark is written in Rust, so no google/nftables I'm afraid.

They could use something like:
    - rustables (v0.8.8, actively maintained) - Rust bindings to libnftnl
     - nftnl (v0.9.2, maintained by Mullvad) - Rust bindings to libnftnl
     - jip-nftables (v0.1.0) - pure-Rust netlink client (no stars)

But the first two are bindings to libnftnl, so IIUC not playing well for the gokrazy usecase.
The third one is pretty new and has no stars.

So not sure how feasible that would be

stapelberg: Ah, OK. Too bad. I don’t know if jip-nftables works well, so let’s not bother podman folks with it.

PR #392, opened 2026-09-16
cmd/ntp: only restore the saved time when it is ahead of the clock

rnavarro:
I run a gokrazy appliance as a KVM guest on Proxmox, and after every reboot the first ten to fifteen seconds of its logs and of the time series it writes carry timestamps from before the reboot, followed by a jump forward. I traced it to `cmd/ntp`: the guest boots with a correct wall clock, and gokrazy steps it backwards to the time saved at the last shutdown.

Here is the boot of 2026-09-15 22:11 PT, taken from the gokrazy web UI. Init starts dhcp, ntp and randomd at 22:11:37, randomd logs at 22:11:38, and then ntp sets the clock back to 22:09:55, which is when the machine shut down about 100 seconds earlier:

```
/gokrazy/dhcp     2026/09/15 22:11:37 gokrazy: attempt 0, starting ["/gokrazy/dhcp" ...]
/gokrazy/ntp      2026/09/15 22:11:37 gokrazy: attempt 0, starting ["/gokrazy/ntp"]
/gokrazy/randomd  2026/09/15 22:11:37 gokrazy: attempt 0, starting ["/gokrazy/randomd"]
/gokrazy/randomd  2026/09/15 22:11:38 randomd.go:38: seeding RNG from /perm/random.seed
/gokrazy/ntp      2026/09/15 22:09:55 ntp.go:52: clock set to 2026-09-15 22:09:55 (from ntp-time-at-last-shutdown)
/gokrazy/dhcp     2026/09/15 22:09:55 dhcp.go:345: waiting indefinitely for eth0 to appear
/gokrazy/ntp      2026/09/15 22:10:06 setting time failed: read udp ...->64.6.144.6:123: i/o timeout
/gokrazy/ntp      2026/09/15 22:11:50 ntp.go:52: clock set to 2026-09-16 05:11:50 UTC (from 0.gokrazy.pool.ntp.org)
```

The 22:11:37 and 22:11:38 lines are the real time: NTP lands on 22:11:50 twelve seconds later. So the kernel handed gokrazy a correct clock, `loadTime` moved it 104 seconds into the past, and everything my program logged or wrote in the next twelve seconds is filed under the previous process's last seconds.

`loadTime` calls `setTimeOfDay` with the saved value with no comparison against the current clock, so this happens on every machine that has a working clock at boot: a PC, a VM, a Raspberry Pi with an RTC hat. It also makes the doc comment in `wait.go` inaccurate for those machines, which says `WaitForClock` is for devices "which lose the time when power is lost and boot with the clock set to January 1, 1970" and that "on machines that boot with a correct time, such as most PCs, this is a no-op". `WaitForClock` does return immediately, but the clock it returns on is now the restored shutdown time rather than the correct one.

This patch restores the saved time only when it is ahead of the current clock. The four cases:

- Boots at the epoch, no battery-backed clock: the saved time is later, so it is still restored. This is the case the restore was added for in #359 and it is unchanged.
- Boots with a correct time: the saved time is behind, so the clock is left alone and NTP refines it. This is my case.
- Boots behind but not at the epoch, for example a slow-drifting RTC: the saved time is later, so it is restored, which moves the clock closer to the truth.
- Boots ahead, for example a fast-drifting RTC: the saved time is behind, so the clock is left alone and NTP corrects it within seconds.

I checked both branches by hand, running the built binary as an unprivileged user with a timefile handed to it on fd 3, where the second case fails at the syscall because the process has no `CAP_SYS_TIME`:

```
# saved time one hour in the past
ntp.go:145: not setting clock to 2026-09-16 04:53:18 +0000 UTC (from ntp-time-at-last-shutdown): system clock (2026-09-15 22:53:18.746729791 -0700 PDT) is already at or ahead of it

# saved time one hour in the future
ntp.go:223: loadTime: setTimeOfDay: syscall.Settimeofday({1789541598 0}): operation not permitted
```

Separately, and not part of this patch: the NTP loop starts at the same moment as the DHCP client, so on my machine its first five queries fail with `lookup N.gokrazy.pool.ntp.org on [::1]:53: connection refused` before `/etc/resolv.conf` exists. That only affects how quickly the first sync lands, and with this patch the clock is correct in the meantime, but it may be worth a `WaitFor("net-route")` or a separate backoff for DNS failures. Happy to send that as its own change if you want it.

Created with help by Claude Opus 5. I reviewed and understood the change, verified the behaviour on my own machine, and am submitting it under BSD-3-Clause.


stapelberg (approved):

stapelberg:
Thanks!

